PR #2, opened 2023-02-15
add auth0 authentication and bcc login 

lukaspoloki:


src/modules/auth0.ts
```diff
@@ -0,0 +1,19 @@
...
+  if (config.auth0) {
+    app.use(
+      createAuth0({
+        domain: config.auth0.domain,
+        client_id: config.auth0.client_id,
+        redirect_uri: window.location.origin,
+        cacheLocation: 'localstorage',
+        audience: config.auth0.audience,
+        useRefreshTokens: true,
+        scope: 'church',
```
SimonSimCity, line 15: Do we need this scope ..?

vite.config.ts
```diff
@@ -3,7 +3,9 @@ import vue from '@vitejs/plugin-vue'
 
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [vue()],
+  plugins: [
```
SimonSimCity, line 6: Why was this changed ..?

tsconfig.json
```diff
@@ -11,7 +11,10 @@
     "esModuleInterop": true,
     "lib": ["ESNext", "DOM"],
     "skipLibCheck": true,
-    "noEmit": true
+    "noEmit": true,
+    "paths": {
+      "~/*": ["./src/*"]
```
SimonSimCity, line 16: BCC Media here already uses `@`, which I would also stick to here.

src/utils/Guid.ts
```diff
@@ -0,0 +1,79 @@
+import { isObject } from '@vueuse/core'
```
SimonSimCity, line 1: What's the reason for including this much logic here? To me, there's not much value to having this and it rather makes me feel like I'm in a C# or Java project ...

src/stores/userStore.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const isUserAuthenticated = computed(() => {
+    return isAuthenticated.value
+  })
+  const currentUser = computed(() => {
+    return user.value
+  })
+
+  const handleLogin = () => {
+    loginWithRedirect({ appState: { targetUrl: '/' } })
```
SimonSimCity, line 20: Does this mean, I'll always end up at the home-page after I logged in or is this setting for something different ..?

src/config.json
```diff
@@ -0,0 +1,7 @@
+{
```
SimonSimCity, line 1: Why not use [`.env` files](https://vitejs.dev/guide/env-and-mode.html#env-files) for configuration?

package.json
```diff
@@ -9,6 +9,10 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "@auth0/auth0-vue": "^2.0.1",
+    "@vueuse/core": "^9.12.0",
+    "pinia": "^2.0.30",
```
SimonSimCity, line 14: I'd argue we do not even need `pinia` after all for this - at least that's what I have the impression of when taking a look at the code bcc media uses (https://github.com/bcc-code/brunstadtv/blob/c111d9f5222edf03b035d91b13231b2a037b0872/web/src/services/auth.ts)

SimonSimCity (changes requested):

package.json
lukaspoloki: for this we could just use the auth0, but why not start with pinia already and have the userStore where we can build some additional logic and things around auth0 as well? we do it in other apps and it works pretty nice and is handy

lukaspoloki: btw it's still a draft PR 😄

PR #3, opened 2023-02-18
Add routing

SimonSimCity:
Added vue-router and relative imports. Relative imports are preferable because you won't have to change imports when moving files around.


kkuepper (approved):
looks fine to me

steffanhalv (approved):

PR #4, opened 2023-02-18
Added linters

SimonSimCity:
This PR includes changes which are part of #3 - please review and merge it before reviewing this one.

Here I've added a command `pnpm run lint`, which will verify that the codebase is according to various rules and style-guides. When running `pnpm run dev`, those errors will stop you from being able to work on, and they are chosen in a way where you either are programming recklessly or you have something which you will stumble on some time later.


Ghostbird:
I'd advocate adding Prettier too, but [probably in a separate PR](https://prettier.io/docs/en/comparison.html)

kkuepper (approved):

SimonSimCity:
@Ghostbird prettier is integrated as a part of eslint (see eslint-plugin-prettier-vue) - but I'm reevaluating if I should remove the package and include `prettier` as a separate tool ...

steffanhalv:
Having a lot of dependencies can cause trouble in the future.

Is this needed?
```
    "eslint-config-airbnb-base": "^15.0.0",
    "eslint-config-airbnb-typescript": "^17.0.0",
```

SimonSimCity:
@steffanhalv can you please elaborate a bit more on that ..? Those packages are needed for the configuration for the linter I have put in place.

package.json
```diff
@@ -6,16 +6,38 @@
...
+    "@typescript-eslint/eslint-plugin": "^5.52.0",
+    "@typescript-eslint/parser": "^5.52.0",
     "@vitejs/plugin-vue": "^4.0.0",
+    "eslint": "^8.34.0",
+    "eslint-config-airbnb-base": "^15.0.0",
+    "eslint-config-airbnb-typescript": "^17.0.0",
+    "eslint-config-prettier": "^8.6.0",
+    "eslint-import-resolver-alias": "^1.1.2",
+    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-promise": "^6.1.1",
```
Ghostbird, line 29: At work we use [`"eslint-plugin-prettier"`](https://www.npmjs.com/package/eslint-plugin-prettier) to integrate ESLint and Prettier. I think correctly set-up it disables all (potentially conflicting) ESLint formatting rules, and instead enforces that the code is formatted the Prettier way.

SimonSimCity: You're right in that `eslint-plugin-prettier` integrates ESLint with Prettier, but it doesn't change any rules - that's what `eslint-config-prettier` is for.

I chose to separate Prettier and ESLint (see 5a047e35bd9318e6e992f002242af478d05c9aa5)

SimonSimCity:
Seems like we're running into some trouble with the tsconfig option `exactOptionalPropertyTypes` 😟

SimonSimCity:
Bugs currently awaiting a resolution:
https://github.com/OpenAPITools/openapi-generator/issues/14792 (worked around by disabling strict: https://github.com/bcc-code/bmm-web/commit/376e1e8bc169acaf501bfbe2cdcc6043e5cd2dce)
https://github.com/OpenAPITools/openapi-generator/issues/14793

Since both are by code generated by OpenAPITools, I'm tempted to temporarily disable the typescript linter for the folder ...

Ghostbird:
> Bugs currently awaiting a resolution: [OpenAPITools/openapi-generator#14792](https://github.com/OpenAPITools/openapi-generator/issues/14792) (worked around by disabling strict: [376e1e8](https://github.com/bcc-code/bmm-web/commit/376e1e8bc169acaf501bfbe2cdcc6043e5cd2dce)) [OpenAPITools/openapi-generator#14793](https://github.com/OpenAPITools/openapi-generator/issues/14793)
> 
> Since both are by code generated by OpenAPITools, I'm tempted to temporarily disable the typescript linter for the folder ...

Does the issue persist if you use OpenAPI v3.1 instead of v3.0.1? They've changed some things around `nullable` handling in the newer one. I think you'll have to use:

```json
"id": { type: ["null", "string"] }
```

Instead of
```json
"id": {
  "type": "string",
  "nullable: true,
}
```

SimonSimCity:
@Ghostbird it's not about the specs, its about how the code generated by the generator works.

Ghostbird:
> @Ghostbird it's not about the specs, its about how the code generated by the generator works.

I know that it generates the wrong code. I was just wondering whether using the newer specs would fix it. Depending on the level of abstraction between parser and code-generator, it might make no difference, or it might generate the code differently. If implemented _properly_ it should make no difference.

steffanhalv:
> Can we merge this PR? It's been open way too long already.

Yes, I can merge it. I'll try to work on this project every Wednesday evening starting this week.

SimonSimCity:
I'll create a new PR for the two outstanding rules. Please enable them in the meanwhile all files other than the generated ones, that would help a lot for my update.

Issue #5, opened 2023-02-20
openapi-generator-cli generate should not need Java Runtime


steffanhalv:
Running this command `npm run openapi-generator-cli` complains about missing Java Runtime:

```bash
Error: The operation couldn’t be completed. Unable to locate a Java Runtime.
Please visit http://www.java.com for information on installing Java.
```

A Java Runtime should not be needed in a Node project @SimonSimCity

SimonSimCity:
That the package `openapi-generator-cli` needs a java runtime is not part of this project and should therefore not be discussed here. In fact, the documentation there clearly outlines the fact that its needed.

Whether we need `openapi-generator-cli` as a dependency, on the other side, is debatable. By removing it as a strict dependency, the requirement of having the java runtime installed would also be removed.

One option would be to discard the generator, which I would not be a fan of. Another, to me more favourable option would be to move the generated code and the generator into a separate repository - like `@bcc-code\bmm-sdk` or something.

steffanhalv:
I think we should make it as easy as possible to get the developer environment up and running.

Moving it to another repository would be a good idea or another option is to remove the `generated` folder from `.gitignore` and add the files to the repository. Then anyone can use the already generated files and we can generate new ones when the BMM API gets an update.

SimonSimCity:
> another option is to remove the generated folder from .gitignore and add the files to the repository. [...] Then anyone can use the already generated files and we can generate new ones when the BMM API gets an update.

This IMO doesn't solve but just postpone the problem you're referring to...

For the other option I need some time and expertise on how to deal with typescript code in npm-packages.

jacobe:
+1 for moving the generated client code to a separate repo and npm package

jacobe:
But until then, having the generated files checked-in would be handy for now

SimonSimCity:
Just for documentation - https://github.com/drwpow/openapi-typescript was mentioned as an alternative, but for my taste, it is too verbose in comparison to the generator we have in place. The generator lets you use a function to execute a request, whereby here you'd have to construct the ur, you'd also yourself have to know which url has which type as response, which is not what I want.

I'll be working on integrating it as npm package today.

steffanhalv:
npm package is nice @SimonSimCity 👍🏼 If you soon got it ready we can wait, but if it taking a while we just check-in the generated files.

SimonSimCity:
Fixed in #13

PR #13, opened 2023-03-08
Replace OpenAPI tools by npm package

SimonSimCity:
In this PR I've replaced the OpenAPI tools (#5) by an npm-package. This allowed us to add back the typescript rules for stricter typing.

I've also added some help like ignoring the `dist` folder on both eslint and prettier, added ts-type-checking to the npm command `lint` and added it to our ci.


src/utils/filters.ts
```diff
@@ -1,11 +1,11 @@
...
-        url.searchParams.set("auth", import.meta.env.VITE_BMM_TOKEN);
-        return url.toString();
+  authorizedUrl(urlString: string | null | undefined) {
+    if (!urlString) {
+      return "";
     }
-}
+
+    const url = new URL(urlString);
+    url.searchParams.set("auth", import.meta.env.VITE_BMM_TOKEN);
```
kkuepper, line 8: that's not the latest version. it should be `localStorage.getItem("token")`

kkuepper: Why does github show the wrong original version?

kkuepper: @SimonSimCity Can you try syncing your fork?
Also, in the future you should just create a branch in bcc-code/bmm-web instead of using the fork.

kkuepper (changes requested):

src/utils/filters.ts
SimonSimCity: Sure, but what you're pointing out here will remain...

kkuepper: The file should look like this. I don't get why it's different in your PR.
https://github.com/bcc-code/bmm-web/blob/main/src/utils/filters.ts

kkuepper (approved):

PR #17, opened 2023-03-08
Auth0 authentication setup

steffanhalv:


SimonSimCity (commented):
Please do not use `npm` in the future, but `pnpm`. Also, when changing eslint rules, I would like to know why.

Would also be nice to know why you renamed the env variable `VITE_BMM_TOKEN` to `BMM_TOKEN` of which I only find references in your code, but not in the `.env`-file...

steffanhalv:
Btw; this was just a fast setup to get the OAuth0 authentication logic up and running, I will rewrite it later and add authentication to a beforehook in the router with a cleaner way to fetch the token for the requests.

SimonSimCity:
> I'll stop pushing the package-lock.json file ;)

Please note that this will not fix the problem. Either we agree on using `pnpm` or we agree on using `npm` as the main package manager, but if each is sticking to his/her preference will get quite tedious IMO. Specially if the person who uses the alternative adds a package like you did in this PR...

steffanhalv:
Thats ok, im fine about using pnpm and I understand the trouble by people stickng to package managers.

But if eslint stops my code to work because

PR #20, opened 2023-03-08
Added tailwind

u12206050:
Add basic store
Add authenticated to store


SimonSimCity:
Would be nice to have the additional things you did (additional to just adding tailwind) in a separate commit, or at least have it mentioned in its first line of the commit message. Otherwise a blame-view can get pretty messy leaving a person quite puzzled.

Issue #11, opened 2023-03-07
How to contribute


kkuepper:
### How do I contribute?
- Assign yourself one of the issues
- fork the repository
- do the work
- create a Pull Request

Some notes
- PR's will be reviewed within 2 business days
- If you don't answer within 2 business days, we might just do needed changes ourselves or do the task ourselves
- the issues are just suggestions what to work on and to make it visible what's already being worked on. If you already know how to contribute, go ahead
- PR's should ideally be merged within a day and not live longer than 7 days
- If you contribute more regularly, we will give you access to the repo, so that you don't need to work on a fork.

SimonSimCity:
As of the example of #2, I think it would be more fair to inform people before taking action and giving them an option to respond. Maybe adding the stale-bot (https://github.com/probot/stale) is a good option?

Joaveinz:
I see this issue has the `documentation` label.

Can we have some documentation on how Tailwind is supposed to be used? From experience, I just write a lot of CSS, so it's not so easy to switch.

It would also be nice to know how strict we are using tailwind and how we can extend it to match our Figma designs.

@SimonSimCity @kkuepper

Joaveinz:
For example, in this case I will need to either use a box-shadow that is part of tailwind, or extend tailwind somehow.

![image](https://github.com/bcc-code/bmm-web/assets/54300414/c6ae67ae-0d73-4bed-b711-5f27775a31d5)

Joaveinz:
Ah, ok. That's fine. My main question was if there were already some unwritten rules regarding this. I can go ahead and write some documentation then, which would allow me to continue :-)

kkuepper:
Deleted some comments to make this less confusing for new developers

sifferhans:
@starkris51 API docs are available [here](https://bmm-api.brunstad.org/documentation/index.html)
With the assignment, you can just comment on which issues you want to contribute to and I can assign you 😁

starkris51:
could you open a discussion tab for this repo, i want to push more open source contributions and discuss community issues

sifferhans:
@kkuepper do you agree to open the discussions tab? I see no problem with it, other than it might not be used that much

starkris51:
> [@kkuepper](https://github.com/kkuepper) do you agree to open the discussions tab? I see no problem with it, other than it might not be used that much

i want to use it to discuss to improve the open source contributions experience, and discuss new tasks to do for the bmm web.

PR #25, opened 2023-03-15
Added testing

SimonSimCity:
Added some testing to the mix 🎉 

For VSCode I've marked the extension for running `vitest` as recommended to make it easier to run individual tests and debug them without the need of any further configuration.

An open discussion is that I'd like to have a coverage-report straight in every PR, which could be solved by adding https://github.com/codecov/codecov-action ... Will be added as soon as a decision is set if this PR will not merged before.


SimonSimCity:
> The code coverage is nice, but you need to specifically look for it.

Got it working 🎉 Its now reporting it on every PR. After we got this merged, it will also do it initially comparing the PR to `main`.

SimonSimCity:
> This PR enables us to write tests, but that doesn't mean that everybody knows how to write useful tests. I've seen it many times that unit tests were started but they turned out to be brittle and tedious and eventually everybody ignored them.
> So I think the more experienced programmers, that know how to write good tests, need to help the less experienced programmers to write useful tests. And they need to bring a lot of patience and empathy and to not get frustrated in the process.

I agree, but it's equally important to note that a senior dev also does good of receiving a code review. I'm all for pair-programming in this regard, but if you're not working at the same time, PRs and code review are the way to go, I think - we should just respect that we need it. This is a challenge for our communication and should go into #11.

Issue #19, opened 2023-03-08
cleanup authentication


kkuepper:
- the first time after logging in, the home screen stays empty since the request has already happened with no token.
- @steffanhalv already has some thoughts on what needs to be done

steffanhalv:
It is fixed in #28, but the build fail because of the linter. I already have a discussion with @SimonSimCity on the same PR about the linter.

PR #28, opened 2023-03-17
fix unauthorized requests & start migration to tailwind

steffanhalv:
This also fix that API requests is not called before authentication.

Please notice I disabled "format on save" because this makes the html hard to read.

Ex. the formatter turns this:

```vue
<div
  v-if="something"
  class="w-24 m-2 bg-green-400 text-white"
  @click="action()"
/>
```

Into this
```vue
<div v-if="something" class="w-24 m-2 bg-green-400 text-white" @click="action()"/>
```

That's not so good when using tailwind as classes is one of the main features. Perhaps the formatter can be fixed to keep newlines? @SimonSimCity


SimonSimCity:
@steffanhalv as far as I can see, I don't have a rule for this in prettier ... The rules I can see around line-spaces are https://prettier.io/docs/en/options.html#bracket-line and https://prettier.io/blog/2018/11/07/1.15.0.html#whitespace-sensitive-formatting. I wouldn't set prettier not to touch any of the html whitespacing (as proposed in https://github.com/prettier/prettier/issues/5380), because the only way I see would be to ignore files, which in our example would be all `.vue` files, which means neither the CSS nor the TS in there would be formatted either.

SimonSimCity:
> This also fix that API requests is not called before authentication.

Would be nice to have this `also` in a separate commit. Eventually, most of the documentation will not be derived from PRs but from the commit-message, which currently doesn't have this information.

steffanhalv:
Here is why I think this gets very annoying @SimonSimCity:

vUadNsXHrG

Most of the styles will be set in the class attribute and every time you save to see the changes and 'crossing the line' the autosave function wraps it in one line if it is too short, or wraps in two lines if it is too long. Perhaps this does not look like a big problem, but when we are going to make a lot of components with hot-reload, this gets very frustrating.

kkuepper:
closes #19

kkuepper:
I fixed the code styling warnings and reverted the change in settings.

We need to get this merged, since other people are waiting for the authentication changes. @jacobe already created an alternative solution #34 to solve the same problem. @steffanhalv @SimonSimCity You can keep discussing the auto-formatting issues on this PR and create a new PR once you have come to a solution.

SimonSimCity:
@steffanhalv If it is that annoying, by all means, feel free to disable the auto-formatting on saving for your environment. If you do not have anything against how the format eventually will end up, I'd prefer sticking to using prettier with the given configuration. For this you could disable it and run it manually fixing up all code-styling after all your work is done; preferably before committing. Does this address your pain or did I misunderstand your point?

steffanhalv:
> @steffanhalv If it is that annoying, by all means, feel free to disable the auto-formatting on saving for your environment. If you do not have anything against how the format eventually will end up, I'd prefer sticking to using prettier with the given configuration. For this you could disable it and run it manually fixing up all code-styling after all your work is done; preferably before committing. Does this address your pain or did I misunderstand your point?

That works 👍

PR #29, opened 2023-03-22
Limit GitHub Actions token permissions

jacobe:
This is the result of scanning the repository for IaC and CI/CD vulnerabilities with [checkov](https://checkov.io)

Testing out whether such a tool could make sense to add as a standard.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -1,5 +1,7 @@
 name: Build and deploy
 
+permissions: read-all
+
 on:
   push:
     branches:


jacobe:
Also, since this a public repository I would only enable running workflows after getting approval (that's a setting in GitHub)

.github/workflows/build-and-deploy.yml
```diff
@@ -1,5 +1,7 @@
 name: Build and deploy
 
+permissions: read-all
```
kkuepper, line 3: what does it do?

kkuepper:
> Also, since this a public repository you should only enable running workflows after getting approval (that's a setting in GitHub)

I don't think it's necessary. For PRs from a fork, it checks if it can build and tests are ok but it skips the deployment.
People that we trust get direct access to the repository and PRs from collaborators are also deployed.

.github/workflows/build-and-deploy.yml
jacobe: It limits the permissions that the GITHUB_TOKEN will get when running workflows. By default it's set to allow write access.

kkuepper (approved):

jacobe:
> > Also, since this a public repository you should only enable running workflows after getting approval (that's a setting in GitHub)
> 
> I don't think it's necessary. For PRs from a fork, it checks if it can build and tests are ok but it skips the deployment. People that we trust get direct access to the repository and PRs from collaborators are also deployed.

👍  But I should really read through the [Security Hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) guide... 😮‍💨  but there are automated checks we can add to test if these best practices are applied. Just looking into [OSSF Scorecard](https://github.com/ossf/scorecard) now. I'll get back when I have an overview - there also probably also things that could be enforced across the whole `bcc-code` org.

PR #31, opened 2023-03-22
feat/private playlist view

### LINKED ISSUE #16: load private playlist details
- new route for private playlists: bmm-web.brunstad.org/playlist/private/:id
- new view
- load information from Api (GET /track_collection/{id})
- show information in view

sifferhans:
fixes #16 
- add route for private playlist
- create view for private playlist and load data based on `id` route param

### DIFF
--- a/src/views/Playlist/PrivatePlaylist.vue
+++ b/src/views/Playlist/PrivatePlaylist.vue
@@ -1,0 +1,45 @@
+<script lang="ts" setup>
+import { Configuration, GetTrackCollectionModel, TrackCollectionApi } from '@bcc-code/bmm-sdk-fetch';
+import { computed, ref } from 'vue';
+
+const props = defineProps<{
+	id: string
+}>()
+
+const trackCollection = ref<GetTrackCollectionModel>({})
+const sortedTracks = computed(() => {
+	const { tracks } = trackCollection.value
+	return tracks?.sort((a, b) => {
+		if (!b.order || !a.order) return 0
+		return b.order - a.order
+	})
+})
+
+new TrackCollectionApi(
+	new Configuration({
+		basePath: import.meta.env.VITE_API_URL,
+		headers: {
+			Authorization: `Bearer ${localStorage.getItem("token")}`,
+			"Accept-Language": "nb,en,zxx",
+		},
+	})
+)
+	.trackCollectionIdGet({
+		id: Number(props.id)
+	})
+	.then(collection => {
+		trackCollection.value = collection
+	})
+	.catch(() => { })
+</script>
+
+<template>
+	<h1>{{ trackCollection?.name }}</h1>
+	<table>
+		<ol class="list-decimal list-inside">
+			<li v-for="track in sortedTracks" :key="track.id">
+				{{ track.title }}
+			</li>
+		</ol>
+	</table>
+</template>
--- a/src/views/Playlist/PrivatePlaylist.vue
+++ b/src/views/Playlist/PrivatePlaylist.vue
@@ -34,7 +34,7 @@ new TrackCollectionApi(
   <h1>{{ trackCollection?.name }}</h1>
   <table>
     <ol class="list-decimal list-inside">
-      <li v-for="track in trackCollection.tracks" :key="track.id">
+      <li v-for="track in trackCollection.tracks" :key="`track-${track.id}`">
         {{ track.title }}
       </li>
     </ol>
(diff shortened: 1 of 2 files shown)


sifferhans:
I see we do this kind of data loading in the `PlaylistOverview` component as well:
```ts
new PlaylistApi(
    new Configuration({
        ....
    })
)
```

Should this be moved to a store or something? It can't be good to instantiate new `PlaylistApi`s everytime a component is mounted?

SimonSimCity:
> Should this be moved to a store or something?

I want to create a vue-plugin for providing the configuration. For the API, I see it as of now not critical to create a new instance of the api-classes, as they're quite light-weight - no heavy lifting or caching done there by which I would justify keeping the instance somewhere.

SimonSimCity (approved):

SimonSimCity:
I'll merge it when the CI tasks run through without error.

src/views/Playlist/PrivatePlaylist.vue
```diff
@@ -0,0 +1,45 @@
...
+	new Configuration({
+		basePath: import.meta.env.VITE_API_URL,
+		headers: {
+			Authorization: `Bearer ${localStorage.getItem("token")}`,
+			"Accept-Language": "nb,en,zxx",
+		},
+	})
+)
+	.trackCollectionIdGet({
+		id: Number(props.id)
```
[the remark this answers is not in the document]
sifferhans: I'll think on it:) I know this has been solved before, I'll take a look

src/views/Playlist/PrivatePlaylist.vue
```diff
@@ -0,0 +1,45 @@
+<script lang="ts" setup>
+import { Configuration, GetTrackCollectionModel, TrackCollectionApi } from '@bcc-code/bmm-sdk-fetch';
+import { computed, ref } from 'vue';
+
+const props = defineProps<{
+	id: string
+}>()
+
+const trackCollection = ref<GetTrackCollectionModel>({})
+const sortedTracks = computed(() => {
```
kkuepper, line 10: I don't think we need to sort them. They come in the right order from the server.

kkuepper: Right order in this case means the order the user put them in.
The order property you're using is meant for within an album.

sifferhans:
@kkuepper do you have any idea why the linter doesn't like this? I cannot seem to figure out what it wants :P

kkuepper:
When hitting save it automatically reformats the file with the following changes:

Screenshot 2023-03-23 at 6 23 41 PM

With those changes the linter doesn't complain anymore.
But I'm getting another error. Probably from `vue-tsc` about `<li v-for="track in trackCollection.tracks" :key="track.id">`. Not sure what it means though.

sifferhans:
@kkuepper Shouldn't we have a .prettierrc file or something to make sure code style stays consistent?

SimonSimCity:
If you reset your local project configuration for vs-code, the configuration we have written in the `.vscode` files should apply. With this configuration in place, it should work.

SimonSimCity:
@sifferhans Not having a `.prettierrc` only means that you do not change the settings prettier normally uses. As far as I know, there are no settings you'd enable, only settings you'd change because you don't like the default settings. We have not yet found a good reason to change what prettier offers us out of the box, so we have sticking to their defaults so far.

SimonSimCity:
@kkuepper Haven't tried the codebase, but would like to see what error you're referring to, just in general ...

sifferhans:
@SimonSimCity Ah ok, but doesn't it fall back on the individual developers VSCode settings if no config file is specified?

kkuepper (approved):

SimonSimCity:
Just some feedback to e066b07 ... I guess you did this because you wanted to fix some issues around the `:key` property. The problem here is not that the property `id` is not a string - the problem is, that it might be null or undefined. By prefixing the value, you've ensured that the key for tracking each individual value of this list, is not empty.

This might also help understanding why `:key` should and how it should be used, as I've followed vues advise to lint on the existence of this property. https://vueschool.io/articles/vuejs-tutorials/tips-and-gotchas-for-using-key-with-v-for-in-vue-js-3/

sifferhans:
@SimonSimCity Yeah, thats why I prefixed it with a string value. It isn't optimal, however, as it means there might be multiple items with the key `track-undefined` or even `track-` in the case of a null value.

What would be the best way to do this, in your opinion? Generate a uid for each item?

SimonSimCity:
To me a good way of fixing this would be to look at the API to see if the property at all should be allowed to be `undefined` 😊  As of now, I'd go the path and expect it to always have a value. In the other components, I've used `id || 0` in order to always have some value even if it would be a duplicate. As I've seen, it only seems important if a list changes, which is unlikely to happen as long as we do not have any manipulative form of reactivity. Just wanted to know if there was another reason behind it, looking at the commit message.

PR #37, opened 2023-03-24
Implement wrapper for images 

fredrikved:
Implement a wrapper for images that need a token.

### DIFF
--- a/src/components/ProtectedImage.vue
+++ b/src/components/ProtectedImage.vue
@@ -12,6 +12,7 @@ const props = defineProps<{
 
 const source = ref("");
 
+// TODO: fix this for tests (auth0 isn't in context in tests at this time)
 const { getAccessTokenSilently } = useAuth0() ?? {
   getAccessTokenSilently: async () => "",
 };
--- a/src/components/ProtectedImage.vue
+++ b/src/components/ProtectedImage.vue
@@ -9,11 +9,7 @@ const props = defineProps<{
 
 const source = ref("");
 
-// TODO: fix this for tests (auth0 isn't in context in tests at this time)
-// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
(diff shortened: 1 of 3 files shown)


src/components/ProtectedImage.vue
```diff
@@ -0,0 +1,26 @@
+<script lang="ts" setup>
+import filters from "@/utils/filters";
+import { useAuth0 } from "@auth0/auth0-vue";
+import { onMounted, ref } from "vue";
+
+const props = defineProps<{
+  src: string | null | undefined;
```
SimonSimCity, line 7: Would it be better not to allow `null` or `undefined` here? There won't be an image to show in that case anyways and the template including this one would have to take care about this (either not showing the image or using a placeholder or whatever fits the case).

src/components/ProtectedImage.vue
```diff
@@ -0,0 +1,26 @@
...
+}>();
+
+const source = ref("");
+
+// TODO: fix this for tests (auth0 isn't in context in tests at this time)
```
SimonSimCity, line 12: Should be fairly easy to do using a `spy` as I've done in `PlaylistOverview.spec.ts`

SimonSimCity (changes requested):

SimonSimCity:
@kkuepper any idea why we don't get a code coverage report? 🤔

kkuepper (approved):

src/components/ProtectedImage.vue
fredrikved: I guess implementing the placeholder outside is better. Originally thought of doing that internally in the component.

SimonSimCity:
Thanks for adding the test and fixing the outstanding comments 😃 I'll merge 👍

SimonSimCity (approved):

PR #38, opened 2023-03-24
Remove vite plugins for eslint and stylelint

SimonSimCity:
As types (typescript) are not checked either, it would be weird to have them in but not typescript. The vite team has deliberately left it out and its better to remain consistent here (https://vitejs.dev/guide/features.html#transpile-only).

Another option, which is also provided on the vites website is to introduce type-checking on change. For me, the argument, that the code editor should take care of that is a perfectly valid argument, which also applies to all the eslint and stylelint rules as well. If they're not explicitly getting the code to fail, this is a better way.

For everyone who doesn't have a such rich editor, he is always free to use VSCode, for which this project has a preset for, or run the package-command `lint` once in a while. The commands there can also be each extracted and ran in a watch-mode if desired, for which I haven't taken the time for.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,8 +49,6 @@
     "tailwindcss": "^3.2.7",
     "typescript": "^4.9.3",
     "vite": "^4.0.0",
-    "vite-plugin-eslint2": "^3.2.1",
-    "vite-plugin-stylelint": "^4.2.0",
     "vitest": "^0.29.2",
     "vue-tsc": "^1.0.11"
   }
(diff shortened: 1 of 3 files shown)


SimonSimCity:
@fredrikved mentioned his concerns about the collection of eslint rules, which I'd take up in a different setting. He mentioned splitting coding issues (e.g. https://eslint.vuejs.org/rules/no-template-shadow.html) from code-styling (e.g. https://eslint.org/docs/latest/rules/arrow-body-style) or at least reporting the latter just as warning. I agree with the objection, just that I'd take those one at a time and don't create too much hassle going into every eslint-setting of the presets I've integrated.

kkuepper (commented):
I can't really comment on this. @fredrikved or @steffanhalv should give feedback.

fredrikved (approved):

PR #41, opened 2023-03-24
fix loading playlists & add route for private playlists

kkuepper:


src/App.vue
```diff
@@ -19,7 +19,7 @@ watch(isLoading, async (loading) => {
       <router-link to="/browse">Browse</router-link>
       <router-link to="/search">Search</router-link>
     </nav>
-    <main class="overflow-scroll">
+    <main class="overflow-scroll" style="padding: 20px">
```
sifferhans, line 22: We should probably use tailwind here instead of inline styling. Try adding the `p-5` class instead :)

sifferhans (approved):

src/App.vue
kkuepper: done

Issue #8, opened 2023-03-07
Profile Dropdown & storing settings


kkuepper:
- see [Figma for how it's supposed to look](https://www.figma.com/file/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?node-id=153%3A127572&t=DM37HzaU1nHnnRjR-0).
- The values should be stored in LocalStorage.

Tasks:
- [x] Implement `Feedback`
- [x] Implement `Contact`
- [x] don't implement `About BMM` since it needs to be different. We'll decide exactly how at a later stage (v0.6 or later).
- [x] implement Profile dropdown (Thank you @belljohan3)

Further tasks have been moved into #162

sifferhans:
Should we use something like [Headless UI](https://headlessui.com/vue/menu) for the menu?

belljohan3:
Hi @kkuepper 
Can we do this task with headlessUi as @sifferhans suggested ? Or is there something else ?
Thanks

belljohan3:
@kkuepper i will give a try to pure tailwind first.

PR #44, opened 2023-03-25
feat/toolbar

### LINKED ISSUE #43: Add Toolbar
- Add toolbar that stays fixed at the top
- has button for profile
- has a title that can be controlled from sub pages
- has a back button (see Lists or Track collections in [Figma](https://www.figma.com/file/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?node-id=153-119499&t=wPoh7jyW63RcYqyk-0))
- has a blur for content hidden under it (optional, can be done in a later issue)

Screenshot 2023-03-25 at 9 57 38 AM

Screenshot 2023-03-25 at 10 01 34 AM

sifferhans:
This PR will fix #43


src/router.ts
SimonSimCity, line 1: Do you know how to type meta? I could have a look at it (I think I got it to work in another project), but if you know it, go ahead...

src/components/AppToolbar.vue
```diff
@@ -0,0 +1,29 @@
+<script lang="ts" setup>
+import { ref } from "vue";
+import { useRoute } from "vue-router";
+
+const route = useRoute();
+
+// TODO: replace placeholder profile menu
```
SimonSimCity, line 7: I wouldn't know what to do when reading this todo. Replace it by what ..?

src/App.vue
```diff
@@ -10,34 +11,34 @@ watch(isLoading, async (loading) => {
     await loginWithRedirect();
   }
 });
+
+const navLinks = [
```
SimonSimCity, line 15: We should have a focus on making such extractions better typed.

SimonSimCity (changes requested):

src/router.ts
sifferhans: I haven't done that before, no. I suppose I need to do some kind of type override in a .d.ts file or something?

src/components/AppToolbar.vue
sifferhans: Yeah, kinda vague haha. I just added a basic dropdown placeholder for the profile menu/dropdown. This will be implemented in a separate PR

kkuepper: Is it better to link to the github issue? #8

src/router.ts
SimonSimCity: I found the docs for it: https://router.vuejs.org/guide/advanced/meta.html

I personally would prefer having it in the `router.ts` file, where it's used - but no strong opinion on that from my side.

Sidenote: The docs mention `interfaces`. In case you don't know about the difference in TypeScript between types and interfaces: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

sifferhans: Nice, thanks for the TS reference! :)
We could just put the declaration in `router.ts` for now, no need to complicate things unneccesarily.

SimonSimCity:
#53 encoperates the changes here.

PR #46, opened 2023-03-28
Added first draft of a media-player

SimonSimCity:
This is a first version of how a media-player could look like - just a POC and some ideas of mine.

The player can be tested on the `browse` route, where I've added a new component `temp-playlist`. As the comment in the file states, this is just here for testing purpose and should be removed as soon as we have a proper listing of tracks - therefore I've not taken the effort of building a well defined component there.

I tried to decouple the media-player and the queue, as I see them as separate concepts where the one should not much relation to the other. This way we can have many "players" where only one of them is actively playing, another might be prefetching data.

Would be nice to hear your feedback on it.


src/components/TempPlaylist.vue
```diff
@@ -0,0 +1,58 @@
+<!-- TODO: This component is just for testing the player until we have a proper listing of tracks. Please remove it as better possibilities of using the player are available. -->
```
kkuepper, line 1: can't you use the PlaylistDetails instead?
https://bmm-web.brunstad.org/playlist/curated/79

SimonSimCity: I get a 404-error when opening the link ...

SimonSimCity: Fixed.

SimonSimCity:
Since all this is just a POC as of now, I refrained of implementing it into the private-playlists as well as of now. If you agree that this is a good way to go, I'll start addressing the `TODO`.

kkuepper:
The one thing I stumbled with is the MediaPlayerInjectionKey. I don't quite like the name but I don't have a better idea either. I guess it would be more intuitive if I used Vue a lot.

In general I think it's fine.
@fredrikved @steffanhalv do you have any feedback?

SimonSimCity:
> The one thing I stumbled with is the MediaPlayerInjectionKey.

I've just chosen it for a personal naming preference - not knowing if there is any tendency of naming in the eco-system. Maybe it's quite an OOP-inspired naming 😄 https://devrant.com/rants/502139/oop-programming-be-like

kkuepper:
I was more thinking to just call it MediaPlayer and make other exports private or protected.
But having inject in the name makes it clear that it needs to be injected before it can be used.

SimonSimCity:
Don't know if this would work out well with lazy loading - haven't tried ... Maybe the player would then be a copy in each component we get in via ll 😓

src/App.vue
```diff
@@ -27,6 +28,9 @@ watch(isLoading, async (loading) => {
       <router-view v-if="isAuthenticated" />
     </main>
   </div>
+  <footer>
+    <media-player />
```
kkuepper, line 32: should this have v-if="isAuthenticated"?

src/components/MediaPlayer.vue
```diff
@@ -0,0 +1,24 @@
+<script setup lang="ts">
+import {
+  MediaPlayerInjectionKey,
+  MediaPlayerStatus,
+  MediaPlaylistInjectionKey,
+} from "@/plugins/mediaPlayer";
+import { inject } from "vue";
+
+const { status, play, pause } = inject(MediaPlayerInjectionKey)!;
+const { currentSong } = inject(MediaPlaylistInjectionKey)!;
```
kkuepper, line 10: why is the exclamation mark necessary? What does it do?

src/views/Playlist/CuratedPlaylistDetails.vue
```diff
@@ -14,6 +17,20 @@ const playlistId = Number(props.playlistId);
 const playlist: Ref<PlaylistModel> = ref({});
 const tracks: Ref<TrackModel[]> = ref([]);
 
+const { setCurrentSong } = inject(MediaPlaylistInjectionKey)!;
+// TODO: Please move this into a plugin or something the-like which configures the authorize-url function.
+let token: string = "";
+(async () => {
+  token =
```
kkuepper, line 24: I would have expected that setCurrentSong takes care of that

SimonSimCity: Debatable, but thanks! I'll move it into the player. It seems to make it a lot easier when dealing with playlists.

src/App.vue
SimonSimCity: I'm actually wondering whether we shouldn't replace the whole scaffolding with some loading indicator or so instead of checking it at each module ...

kkuepper (changes requested):

src/views/Playlist/CuratedPlaylistDetails.vue
kkuepper: I think this is the only thing preventing this from being merged

SimonSimCity: Should be fixed now.

kkuepper (approved):

PR #47, opened 2023-03-29
Refactored api configuration into vue-plugin

SimonSimCity:
I was unsure if we should leave the language as header or add another middleware which can be overwritten per request. Just in case we want to allow the user to specifically choose a language. Not sure how this is implemented in the app - would be good if you could give us a hint @kkuepper


kkuepper:
> I was unsure if we should leave the language as header or add another middleware which can be overwritten per request. Just in case we want to allow the user to specifically choose a language. Not sure how this is implemented in the app - would be good if you could give us a hint @kkuepper

It's a super edge case. In 99% it's fine to set the header in the middleware.
https://github.com/bcc-code/bmm-app/blob/44b053a9400198a3f78f57f0f8d01ecea4603c30/BMM.Api/Implementation/Clients/TracksClient.cs#L62 shows how it's done in the App.
I would image we'd need something like that would be nice.

```
new TracksApi()
  .changeLanguage(language)
  .trackGet(id);
```

kkuepper:
Overall, I think it looks good. I agree that the specific files (e.g. playlistApi.ts) aren't really necessary.

One thing I noticed is, that using the swagger is harder than I imagined.
Screenshot 2023-03-29 at 9 18 24 AM

I guess creating a separate file, in part was done to hide that complexity a bit. What can we do to make the generated methods more intuitive?
One option would be to hide admin endpoints (e.g. playlistOverview or playlistPut). I also think putting Get or Post at the start of the name instead of the end would help.

SimonSimCity:
> I would image we'd need something like that would be nice.

Just found out, we can overwrite headers using the last parameter of a request, so this is possible.

kkuepper:
Is this still a draft? Because it's still a draft PR.

kkuepper:
How can we improve the situation? I see 2 options:
- create a better swagger in BMM Api (Karsten)
- change settings of the Code generator (Simon)

The best is probably if we do both. I have to admit that I don't really have much experience with Swagger and would welcome some feedback about how to improve it.

SimonSimCity:
@kkuepper I don't know how much control you have about the swagger file itself and how much is generated from a place where you've little to no control over. In my opinion, a swagger file can be seen as an interface definition, and the first attempt I'd set is if the project rather should go code- or swagger-first. The other should then be generated to ensure compatibility. As far as I understand, you've (consciously or unconsciously) decided to go code-first on the API, which means the outcome (the swagger file) can only be as good as its generator. This means the package for this project is generated code based on generated code 😏 

IMO its up to you to decide whether its worth the hassle, as this might result in a bigger work due to all the dependencies. The focus and main outcome should be defined very clearly though. It doesn't seem you have a lot of hassle with type-safety, as the interface of the API is quite stable and neither the BMM App nor the BMM Suite (just two of the projects apart of this which I know of) are using any type of safety-net, like ensuring types are according to the interface-definition, when accessing the API. We also found out that the interface-definition sometimes didn't match what the system returned ...

kkuepper:
I removed admin endpoints from the current swagger definition and created a new definition that also has those protected endpoints.
That should make it less confusing to find the right one.
From my point of view this PR can be merged as soon as bmm-sdk has been updated.

SimonSimCity:
Seems we now have 6,486 less lines of code in the package 😆 https://github.com/bcc-code/bmm-sdk/commit/17b3f538050cb221506365a63f26a83b324e9522

PR #51, opened 2023-04-03
Introducing translation

SimonSimCity:
What I did so far:

- Installed i18n as the plugin for translation (BTV also uses `@intlify/unplugin-vue-i18n` and I'm not sure of the benefit of this plugin ...)
- Installed `@intlify/eslint-plugin-vue-i18n`, so eslint can report keys not found in translation and other cool things
- The import of languages is set up to have a type-error if the language files do not have the same structure (one of them is missing a translation)

What's open:

**Translation files are not parsed**
By this we are neither able to get a warning if HTML is used in a translation, neither are we notified about superfluous translations - both of which are available as part of the eslint package.


kkuepper:
I'll set up crowdin integration after this is merged. As far as I can tell the json should work fine with it.
Looks fine so far.

PR #53, opened 2023-04-05
Feat/toolbar (follow up on #44)

SimonSimCity:
Continues the changes of #44 and also some additional ones.


sifferhans:
This closes #43

src/locales/de.json
```diff
@@ -2,6 +2,8 @@
   "nav": {
     "home": "Home",
     "browse": "Stöbern",
+    "playlist": "Playliste",
```
kkuepper, line 5: I'm not sure if you can translate it here. These values might be overridden with the english values again.

src/router.ts
```diff
@@ -1,31 +1,63 @@
...
+import i18n from "./plugins/i18n";
+
+declare module "vue-router" {
+  interface RouteMeta {
+    toolbarTitle: Ref<string>;
+  }
+}
+
+const { locale, t } = i18n.global;
+const rT = (translate: () => string) => {
```
kkuepper, line 12: what is this for? I don't understand why it's needed. And maybe rT is not a good name then

SimonSimCity: This makes the translation, which is passed as part of the route, reactive. Its naming is derived from the translation functions (called `t` or `tn`) as `reactiveTranslation`.

Without this addition, a title in the toolbar won't change when switching translation.

kkuepper: would you mind calling the variable reactiveTranslation? Or autoUpdatedTranslation?
t() is fine since you can look at the import and understand it's for translation. Also translation is used everywhere so it's nice to have a very short name. But rT is confusing since it's hard to understand what's the difference to t() and when to use one over the other. Also it's a local function that's only used in this file, so no reason to give it an understandable name.

kkuepper (changes requested):
rT should be renamed & resolve merge conflicts.
After that it can be merged

src/router.ts
SimonSimCity: Also added a static type definition to the function, which makes the return type and the function definition explicit and independent of the implementation.

PR #54, opened 2023-04-06
component-for-list-of-tracks

belljohan3:


src/components/tracksList.vue
```diff
@@ -0,0 +1,46 @@
...
+const trackAlbum = ref("");
+const trackSubtype = ref("");
+
+const { getAccessTokenSilently } = useAuth0();
+
+onMounted(async () => {
+  const token = await getAccessTokenSilently();
+
+  source.value = filters.authorizedUrl(props.src, token);
+  trackNumber.value = filters.authorizedUrl(props.id, token);
```
sifferhans, line 28: What is the reason for using the `authorizedUrl` filter here? :)

belljohan3: I said to myself that an authorization was needed for each piece of data that we want to have. not true ? If I did not write well, can you suggest me something or a correction?

sifferhans: Its a good thought! I think, however, that the `authorizedUrl` util is primarily used to authorize image urls (or URLs in general).

If you look at the code for the `authorizedUrl` util, you'll see it is setting some headers on a `URL` object.
We probably don't need to pass numbers through, as numbers aren't URLs.

belljohan3: Ok ! Thanks - How do you suggest me to do then @sifferhans ?

sifferhans: I think you can simply use the props directly instead of using separate `ref`s.

You can also just use the `ProtectedImage` component instead of an `img` element, which will handle authorization of the image source for you 😊

belljohan3: Okay ! I will try it afterwards !

SimonSimCity: It's actually very rare you'd need the function `authorizedUrl` in your code. Maybe a good advice is to first try a minimalistic approach (less code) and add what's needed - as you understand their usage.

src/views/Playlist/PrivatePlaylist.vue
```diff
@@ -19,13 +21,31 @@ new TrackCollectionApi()
...
+
+<!-- This template is for test purpose. It might not work properly -->
+<template>
+  <h2>{{ trackCollection?.name }}</h2>
+  <div
+    v-for="track in trackCollection.tracks"
+    :key="track.id"
+    class="flex flex-row flex-wrap"
+  >
+    <tracksList
```
SimonSimCity, line 41: Why do you need to dissolve the type `track` here? Wouldn't it be easier to just pass on the track instead of adding a property for each primitive type?

src/components/tracksList.vue
```diff
@@ -1,39 +1,46 @@
...
+<script lang="ts" setup>
+import filters from "@/utils/filters";
+import { useAuth0 } from "@auth0/auth0-vue";
+import { onMounted, ref } from "vue";
 
-import ProtectedImage from "./ProtectedImage.vue";
+// import ProtectedImage from "./ProtectedImage.vue";
 
-const tracks: Ref<TrackModel[]> = ref([]);
+const props = defineProps<{
```
SimonSimCity, line 8: Please be more specific than just `any`. By just making the type `any`, this can be anything. A boolean, array, function, null, etc...

src/views/Playlist/PrivatePlaylist.vue
SimonSimCity: Please give the naming a second chance. What is this of an element? You've called it `tracksList` -but this element doesn't look like anything I'd relate when reading it's tag - or the parameters don't fit...

SimonSimCity (changes requested):

sifferhans:
@belljohan3 You have named the component `tracksList`, but it doesn't look like the component is a list. Would it be better to rename the component to `Track`, since it only displays the data of one track?

Maybe also try passing the whole track as a prop instead of passing separate props for title, image etc?
Example:
```vue
<Track
  v-for="track in tracks"
  :key="track.id"
  :track="track"
/>
```

belljohan3:
I will fix it as soon as possible! @steffanhalv thanks

sifferhans:
> > @belljohan3 You have named the component `tracksList`, but it doesn't look like the component is a list. Would it be better to rename the component to `Track`, since it only displays the data of one track?
> > Maybe also try passing the whole track as a prop instead of passing separate props for title, image etc?
> > Example:
> > ```
> > <Track
> >   v-for="track in tracks"
> >   :key="track.id"
> >   :track="track"
> > />
> > ```
> 
> What would be the type the prop `track` @sifferhans ?

You can get the `TrackModel` type from the BMM SDK package 😊
```ts
import type { TrackModel } from '@bcc-code/bmm-skd-fetch'
```

belljohan3:
@SimonSimCity please can you have a loook at this ? it would be nice if we can merge it as soon as possible

PR #56, opened 2023-04-11
Added a general error page

SimonSimCity:
Not a requirement as of now, but it would be good to have something in this direction to show the user he went astray.

The messages should be revised by someone, but I refused to use `404`, which is technical and the user shouldn't have to be bombarded with protocol-specific error messages, but by a clear natural readable text IMO 😏


src/locales/en.json
```diff
@@ -3,5 +3,11 @@
     "home": "Home",
     "browse": "Browse",
     "search": "Search"
+  },
+  "error": {
+    "page-not-found": {
+      "headline": "Error",
+      "text": "No route for \"{path}\""
```
kkuepper, line 10: Sorry. We can't find the page "{path}".

kkuepper (changes requested):

src/locales/en.json
SimonSimCity: Fixed.

kkuepper (approved):

PR #61, opened 2023-04-14
Fail switching playlist routes

### LINKED ISSUE #60: [BUG] Cannot navigate between private playlists
The RouterView doesn't update when navigating between private playlists.

I suspect it is because the props object isn't reactive. This is how it is implemented in `views/Playlist/PrivatePlaylist.vue`:

```ts
const props = defineProps<{
  id: string
}>()

new TrackCollectionApi()
.trackCollectionIdGet({ id: Number(props.id) })
                                    ^ This is not reactive
//...
```

CleanShot 2023-04-14 at 10 27 01

SimonSimCity:
Fixes #60

### DIFF
--- a/src/components/ProtectedImage.vue
+++ b/src/components/ProtectedImage.vue
@@ -1,7 +1,7 @@
 <script lang="ts" setup>
 import filters from "@/utils/filters";
 import { useAuth0 } from "@auth0/auth0-vue";
-import { onMounted, ref } from "vue";
+import { watch, ref } from "vue";
 
 const props = defineProps<{
   src: string;
@@ -11,11 +11,15 @@ const source = ref("");
 
 const { getAccessTokenSilently } = useAuth0();
 
-onMounted(async () => {
-  const token = await getAccessTokenSilently();
+watch(
+  props,
+  async () => {
+    const token = await getAccessTokenSilently();
 
-  source.value = filters.authorizedUrl(props.src, token);
-});
+    source.value = filters.authorizedUrl(props.src, token);
+  },
+  { immediate: true }
+);
 </script>
 <template>
   <img :src="source" />
--- a/src/components/ProtectedImage.vue
+++ b/src/components/ProtectedImage.vue
@@ -12,7 +12,7 @@ const source = ref("");
 const { getAccessTokenSilently } = useAuth0();
 
 watch(
-  props,
+  () => props.src,
   async () => {
     const token = await getAccessTokenSilently();
 
--- a/src/views/Playlist/CuratedPlaylistDetails.vue
+++ b/src/views/Playlist/CuratedPlaylistDetails.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { ref, Ref, watch } from "vue";
+import { inject, ref, Ref, watch } from "vue";
 import {
   TrackModel,
   PlaylistModel,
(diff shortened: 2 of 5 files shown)


src/components/ProtectedImage.vue
```diff
@@ -11,11 +11,15 @@ const source = ref("");
 
 const { getAccessTokenSilently } = useAuth0();
 
-onMounted(async () => {
-  const token = await getAccessTokenSilently();
+watch(
+  props,
```
sifferhans, line 15: You can probably just watch `props.src` here.
```ts
watch(
  () => props.src,
  async () => { /* ... */ }
)
```

src/views/Playlist/CuratedPlaylistDetails.vue
```diff
@@ -17,18 +17,24 @@ const tracks: Ref<TrackModel[]> = ref([]);
...
-  })
-  .catch(() => {});
-new PlaylistApi()
-  .playlistIdTrackGet({ id: playlistId })
-  .then((list) => {
-    tracks.value = list;
-  })
-  .catch(() => {});
+watch(
+  props,
```
sifferhans, line 21: You can probably just watch `props.playlistId` here.

```ts
watch(
  () => props.playlistId,
  newId => {
    new PlaylistApi()
      .playlistIdGet({ id: newId })
  }
)
```

src/views/Playlist/PrivatePlaylist.vue
```diff
@@ -3,20 +3,26 @@ import {
...
 const trackCollection = ref<GetTrackCollectionModel>({});
 
-new TrackCollectionApi()
-  .trackCollectionIdGet({ id: Number(props.id) })
-  .then((collection) => {
-    trackCollection.value = collection;
-  })
-  .catch(() => {});
+watch(
+  props,
```
sifferhans, line 15: Same here, watch `props.id` instead of the whole props object:)

```ts
watch(
  () => props.id,
  newId => {
    new TrackCollectionApi()
      .trackCollectionIdGet({ id: Number(newId) })
  }
)
```

sifferhans (commented):
Looks good, only try watching specific props if you don't need to watch the entire object :)

SimonSimCity:
> Looks good, only try watching specific props if you don't need to watch the entire object :)

By which I'd have to unwrap the object and create watchable values of those I want to watch. In the components here it doesn't make any difference, so I left it this way.

sifferhans:
You can pass in a getter function instead of the prop directly, which will solve it :) 
https://vuejs.org/guide/essentials/watchers.html#basic-example
```ts
watch(
  () => props.id, // <-- getter function
  (newId) => {
    // ...
  }
)
```

But it's not that important when there's only a single prop, so it should be fine

SimonSimCity:
Ah! Thanks for pointing this out 😃 Wasn't aware of this until now ... I only knew that you could wrap your props into [toRefs](https://vuejs.org/api/reactivity-utilities.html#torefs) (and get reactive properties from there (which you have to do in a separate line, because [`defineProps` is a compiler macro](https://vuejs.org/api/sfc-script-setup.html#defineprops-defineemits)), but using a getter here makes perfect sense! I'll update 😎 

For those staring at the docs unable to find what we're talking about, you need to switch `API Preference` from `Options` to `Composition` at the upper left 😏

SimonSimCity:
Just discovered that my changes would not have helped much for the `CuratedPlaylistDetails` view - fixed now.

sifferhans (approved):
Looks good to me!

PR #62, opened 2023-04-19
Auth0 as plugin

SimonSimCity:
Refactored the implementation of how data provided by auth0 is accessed in the system.


src/components/ProtectedImage.vue
```diff
@@ -1,15 +1,15 @@
...
+import { inject, onMounted, ref } from "vue";
 
 const props = defineProps<{
   src: string;
 }>();
 
 const source = ref("");
 
-const { getAccessTokenSilently } = useAuth0();
+const { getAccessTokenSilently } = inject(AUTH0_INJECTION_KEY)!;
```
kkuepper, line 12: to be honest I think the previous way was easier to understand. What's the benefit of the changes in this PR?

SimonSimCity:
Was meant to use dependency injection instead of straight importing them.

Makes a.o. testing easier because you do not have to go through the trouble of mocking node-modules. This forces us to run tests sequentially and has also quite some other limitations.

kkuepper (approved):
looks fine

PR #76, opened 2023-04-22
hover actions on track

### LINKED ISSUE #70: hover actions on track
- Demo: https://www.figma.com/proto/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?node-id=57-9803&scaling=min-zoom&page-id=153%3A127571
- https://www.figma.com/file/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?node-id=57-9803&t=teCmnYuVybxA7942-0
- Download action
- Add to playlist action

bjornarhagen:
Closes #70 

* Installs https://nuxt.com/modules/icons for Icon handling
* Adds Icon component
* Update TrackItem with hover effect and icons


kkuepper:
@bjornarhagen can you run `pnpm lint` locally and fix the issues that come up?

kkuepper (approved):

components/Icon.vue
sifferhans, line 1: What's the reason for making this wrapper component? Can't we just use the `nuxt-icon` component directly?

components/track/TrackItem.vue
```diff
@@ -1,28 +1,92 @@
...
+        <Icon name="download" filled class="only-show-on-hover text-2xl" />
+        <Icon name="queue" filled class="only-show-on-hover text-2xl" />
+        <Icon name="options" filled class="text-2xl" />
+      </div>
     </div>
   </li>
 </template>
+
+<style>
+@tailwind base;
```
sifferhans, line 48: Using `@apply` is not recommended by TailwindCSS, as it kinda defeats the purpose of using Tailwind in the first case.

- Try using a drop shadow for the hover styling
- use the `group` and `group-hover:` classes to style children based on the parents hover state

sifferhans (changes requested):
Left some comments for what should be changed 😊

components/track/TrackItem.vue
bjornarhagen: Done ✅

components/Icon.vue
bjornarhagen: For some reason it rendered in the incorrect order for me, but worked when I wrapped it 🤷

bjornarhagen:
nuxi typecheck seems to be failing to due the nuxt-icons dependency. I'm unsure how to fix that

sifferhans (approved):

.gitignore
```diff
@@ -14,3 +14,6 @@ cypress/videos/*
 .vscode/*
 !.vscode/extensions.json
 !.vscode/settings.json
+
+# OS files
+.DS_Store
```
SimonSimCity, line 19: This should be done in your global `.gitignore` file instead of part of this repo, because its part of the environment you have.

components/IconComponent.vue
```diff
@@ -0,0 +1,68 @@
+<script setup lang="ts">
```
SimonSimCity, line 1: Please add a comment here that this is a copy from another package (also mention the name of the package 🤗 ), since the code isn't ours. The license of the package doesn't require it, but it would be good for us in terms of reliability.

components/track/TrackItem.vue
```diff
@@ -8,21 +8,60 @@ defineProps<{
...
+          />
+        </div>
+      </div>
+      <div class="mr-auto w-1/3">
+        <h4 class="font-semibold">{{ track.meta?.title }}</h4>
+        <span class="text-slate-700" v-if="track.meta?.artist">
+          {{ track.meta?.artist }}
+        </span>
+      </div>
+      <div
```
SimonSimCity, line 42: Why not use `v-else` here? https://vuejs.org/guide/essentials/conditional.html#v-else

SimonSimCity (changes requested):

components/track/TrackItem.vue
kkuepper: they are quite far apart. I think it's more readable not to use it.

SimonSimCity (approved):

PR #88, opened 2023-04-23
Typed routes

SimonSimCity:
This PR also sets how i18n should behave in regards to routing.


sifferhans (approved):

sifferhans:
@SimonSimCity It seems the typed-router package is updated with the i18n fixes 
https://github.com/victorgarciaesgi/nuxt-typed-router/releases/tag/v3.2.0

SimonSimCity:
I expect the error we had here to be a side effect of #86, in which we removed the fix on `vue-tsc` version. By this the type safety was not guaranteed to be checked in the same version. Don't know what this means for the other libs ...

SimonSimCity:
@sifferhans I'll update, but this shouldn't affect us ... https://github.com/victorgarciaesgi/nuxt-typed-router/issues/87#issuecomment-1518996773

SimonSimCity:
The problem about `vue-tsc` was not anything related to peer-dependencies - it was simply a package which was not installed. When then `npx` tried to run it, it asked to install it instead of which the version was different on what I used locally. I've added it as a dev-dependency.

Our handling of peer-dependencies is, as pointed out in #86, completely valid in that we let the packaging system figure out a version fitting to all peers. This is reflected in the `pnpm-lock.yaml` file which keeps track of all this.

components/DropdownMenu.vue
```diff
@@ -47,14 +49,14 @@ function menuItemClick(event: Event, item: DropdownMenuItem) {
         <p
           v-else-if="item.clickFunction"
           class="flex justify-start items-center gap-1 py-2 px-3"
-          @click="(event) => menuItemClick(event, item)"
+          @click="(event: MouseEvent) => menuItemClick(event, item)"
```
SimonSimCity, line 52: Types don't have to be defined after updating `typescript `to v5, but then we run into other type-problems in `components/PageHeading.vue` 😰

PR #96, opened 2023-04-23
Feat/67 albums subalbums

### LINKED ISSUE #67: album & subalbums
- https://www.figma.com/proto/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?node-id=172-13920

linesharina:
Closes #67


components/sidebar/SidebarElement.vue
```diff
@@ -3,6 +3,7 @@ const links: Record<"title" | "url", string>[] = [
   { title: "Home", url: "/" },
   { title: "Browse", url: "/browse" },
   { title: "Search", url: "/search" },
+  { title: "Album", url: "/album/91530" },
```
linesharina, line 6: This should be changed once we have a page for viewing all albums

SimonSimCity: Could be resolved by #97

pages/album/[id].vue
```diff
@@ -0,0 +1,48 @@
+<script lang="ts" setup>
+const { t } = useI18n();
+toolbarTitleStore().setReactiveToolbarTitle(() => t("nav.album"));
```
Comment (SimonSimCity) [inline pages/album/[id].vue:3]: Maybe someone else will have to answer this ... Wouldn't it make sense to have the name of the album as title in the toolbar?

pages/album/[id].vue
```diff
@@ -0,0 +1,48 @@
...
+
+const { id } = useRoute().params;
+const albumId = Number(id);
+
+const { data: album } = useAlbum({ id: albumId });
```
Comment (SimonSimCity) [inline pages/album/[id].vue:8]: When now navigating from one album to another, the view won't update. We've had this before in #60 with playlists.

SimonSimCity (changes requested):
@sclausendk, @kkuepper I just realized that this implementation (maybe the design as well) only allows for albums in 2 nested levels. I don't know if we have more, but as of the structural design, the API allows unlimited nesting of albums, if I'm not mistaken...

@linesharina Don't hesitate to contact me about what I wrote (whether its unclear, you don't agree, want me to explain more or assist, ...)

kkuepper:
> @sclausendk, @kkuepper I just realized that this implementation (maybe the design as well) only allows for albums in 2 nested levels. I don't know if we have more, but as of the structural design, the API allows unlimited nesting of albums, if I'm not mistaken...

Theoretically that's allowed, but the API also only returns up to 2. We haven't needed more in 15 years, so I think it's fine.

pages/album/[id].vue
Comment (sifferhans) [reply pages/album/[id].vue:8]: I think it might work to wrap the `Number(id)` in a `computed`, such as this:

```ts
const albumId = computed(() => Number(id))
```

Comment (SimonSimCity) [reply pages/album/[id].vue:8]: I highly doubt this, because `id` is of type `string`. Having a string inside a `computed` function won't do anything, it must have a form of reactiveness, like `Ref<T>`.

What I still need to wrap my head around is, that nuxt changes something on the routing so navigating to the same route triggers a re-rendering of the page; that said, my changes from #60 are not required anymore, but we are constantly re-rendering components even though we don't need it ... weird ...

kkuepper:
@linesharina if you don’t mind I’ll take a look at the remaining issues tomorrow

components/album/SubAlbum.vue
```diff
@@ -0,0 +1,61 @@
+<script lang="ts" setup>
+const props = defineProps<{
+  id?: number;
+  active: boolean;
+}>();
+
+if (!props.id) {
```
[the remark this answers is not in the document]
linesharina: Hey I just followed the AlbumModel from the API where the id is set as a nullable property

components/album/SubAlbum.vue
```diff
@@ -0,0 +1,61 @@
...
+      :class="active ? 'active max-h-fit relative' : 'h-0 overflow-hidden'"
+    >
+      <div v-for="(track, i) in album?.children" :key="i">
+        <p class="p-2 hover:bg-slate-100 rounded-md">
+          {{ track.id }}
+        </p>
+      </div>
+    </section>
+  </section>
+  <!-- <TrackList :tracks="album.chldren" /> -->
```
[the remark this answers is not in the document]
linesharina: Just an oversight on my end, removed it now

pages/album/[id].vue
Comment (SimonSimCity) [reply pages/album/[id].vue:3]: Just for clarification - this sets the red, not the blue text. The blue text is fixed as of today. Screenshot 2023-04-27 at 10 20 36

Comment (SimonSimCity) [reply pages/album/[id].vue:3]: I'd propose to only take the name of the album. Many websites have a breadcrumb menu, but this might not be valuable for our website, because we do not have such a nested structure of websites that this could be useful 😁

>We could do it for playlists as well.

I agree 😎

components/album/SubAlbum.vue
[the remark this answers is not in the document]
SimonSimCity: But why not handle this in the outer component which uses this? Why allow data to pass in to this component, which we don't know how to handle here? 😏

PR #99, opened 2023-04-26
Added empty prettier config

### LINKED ISSUE #90: setup a formatting config
Currently the formatting is handled by Prettier.
But we do not have any config file that defines what the formatting should be, so it uses the default.

The problem with this is that developers might have changed their default. Also the default might be different between different versions of Prettier.

If we setup a .prettierrc file in the root folder, Prettier will use that instead of the default.

Also a suggestion, the default line width of 80 is very short, we should set it to 120-140, the HTML/Vue template becomes much more readable.

SimonSimCity:
Will help to ignore the config of vscode people might have in place (https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode#visual-studio-code-settings)

Fixes #90


SimonSimCity:
@sifferhans can you please check if this helps your setup and merge if you can confirm it does?

sifferhans:
@SimonSimCity Seems to work for me

Issue #33, opened 2023-03-23
Define error handling strategy


SimonSimCity:
Currently we disregard all errors coming from a request to the API. An ideal outcome of this ticket would be a definition of how basic errors (e.g. server couldn't be reached/network down) could be handled - some kind of universal solution, which then would be implemented in all existing cases of failure and should be followed by all subsequent PRs.

SimonSimCity:
I see 3 categories of errors, which should be handled and logged differently:

1. User relevant errors - the user has done something wrong. These errors should be shown to the user. Logging is not that relevant, because its nothing we should deal with, but maybe good to have them logged somewhere to improve UX by updating user-flows to avoid them. I don't see many cases for this being possible, because we only have limited space for user-input.
2. Errors within this application (4-axis https://youtu.be/AdMqCUhvRz8?list=FL5t9y4K3qdsL0nSRtzhaoIw&t=1201 - 4 min). Fatal errors will mostly be Electron internal, but the non-fatal ones are ours to deal with, which I wouldn't care much about as of now. Explicit ones can be logged, and for implicit ones monitoring system resources and user response can help.
3. Errors in the API. While also part of non-fatal errors (explicit or implicit), both should IMO not be logged by this instance, but by the API. This includes an API not responding (could be our problem - API is down - or the users - network is disconnected) or returning a 5xx error-code.

**A plan for the strategy of handling all errors:**
For 1st we will need a way to report those errors, which should be discussed with @sclausendk. For 2nd and 3rd, we need a system to centrally report errors and exceptions but also to monitor abnormal system behavior. Logs of the 2nd category should be logged distinguishable from the 3rd. A set of developers should have access to the mentioned system and within a given time-frame (e.g. each month) have a look at the errors logged there and define action-items for handling them.

Development-tasks I see to fulfill the plan above:
1. Decide on a system for reporting errors and logs and implementing it for Web and Electron.
2. (low) Find out if there's a way of logging and reporting fatal errors (e.g. [MacOS has crash-reports](https://developer.apple.com/documentation/xcode/acquiring-crash-reports-and-diagnostic-logs#Collect-crash-reports-from-TestFlight-and-the-App-Store))
3. (low) Define which system-metrics and user-actions to log to supervise implicit, non-fatal errors.

**Next Steps:**
This is just a set of ideas I have. @kkuepper Would be nice to get your opinion on this.

sifferhans:
[...]

PR #101, opened 2023-04-29
profile menu started

belljohan3:


components/AppToolbar.vue
```diff
@@ -1,9 +1,14 @@
+<!-- eslint-disable @intlify/vue-i18n/no-raw-text -->
 <template>
   <header
-    class="py-4 px-6 border-b border-slate-100 bg-white/80 backdrop-blur-md sticky top-0"
+    class="flex justify-between py-4 px-6 border-b border-slate-100 bg-white/80 backdrop-blur-md sticky top-0"
   >
     <strong v-if="toolbarTitleStore().toolbarTitle">
       {{ toolbarTitleStore().toolbarTitle }}
     </strong>
+    <light class="flex">
```
sifferhans, line 9: I dont think there is a HTML element called `light` :) Did you want the text to be a light font? If so, use a span instead with the class `font-light`

components/DropdownMenu.vue
```diff
@@ -56,14 +56,14 @@ function menuItemClick(event: Event, item: DropdownMenuItem) {
         </p>
         <p
           v-else
-          @click="(event: MouseEvent) => event.stopPropagation()"
           class="flex justify-start items-center gap-1 py-2 px-3"
+          @click="(event: MouseEvent) => event.stopPropagation()"
```
sifferhans, line 60: What is the `stopPropagation` for?

components/DropdownMenu.vue
```diff
@@ -56,14 +56,14 @@ function menuItemClick(event: Event, item: DropdownMenuItem) {
...
         >
           <IconComponent v-if="item.icon" :name="item.icon" />
           <span>{{ item.text }}</span>
         </p>
       </li>
     </ul>
-    <div @click="close" class="fixed inset-0 z-10"></div>
+    <div class="fixed inset-0 z-10" @click="close"></div>
```
sifferhans, line 67: Can we use a button here instead of a div? Since we have a click handler here it makes sense to use the semantic HTML element

layouts/default.vue
```diff
@@ -1,9 +1,7 @@
 <script lang="ts" setup>
 import { MediaPlaylistInjectionKey } from "~/plugins/3.mediaPlayer";
 
-const onError = (error: any) => {
-  console.error(error);
-};
+const onError = (error: any) => error;
```
sifferhans, line 4: What is the reason for this change? 
With this change the `onError` function only returns the error we get from the `NuxtErrorBoundary`, but we dont do anything with it

sifferhans (changes requested):

components/DropdownMenu.vue
belljohan3: I don't remember that i made changes in this file.

layouts/default.vue
belljohan3: I made these changes to avoid getting the workflow errors! when I do the `git push`

components/DropdownMenu.vue
belljohan3: I think it is @bjornarhagen who worked on that !!

SimonSimCity:
@sifferhans IMO some of your comments should have gone to a different PR. Most of the changes here were sorely done to reduce the amount of warnings eslint spits out, mostly https://eslint.vuejs.org/rules/attributes-order.html.

Because you asked: The event propagation is needed for the `click` event not to bubble up to the `track` item, which the `dropdownMenu` component is a part of, which in turn has a different `click` event. See https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation for further information. IMO, the function `stopPropagation()` shouldn't be called inside a function, because its then hidden from the HTML-structure because of which it is called. IMO, we should take a turn on all events and call `event.stopPropagation(); myFunction(event)` if possible, to know where the propagation is done and where its not done. Sometimes its also unwanted to stop the propagation - and in those cases its better to have the code where its important to know about the side-effects. But this is more of me going into the nature of functional programming. Comments welcome 🤗

components/AppToolbar.vue
```diff
@@ -1,9 +1,14 @@
+<!-- eslint-disable @intlify/vue-i18n/no-raw-text -->
```
SimonSimCity, line 1: I think you added this because eslint started to complain about something in this component. Please avoid using disabling of rules, if its not strictly wanted to leave them out. This rule should just create a warning, no error, so you should be able to deploy just fine.

To us, having a warning here, shows that there's still some work to do - in your case, there's text which should be translated. By having this eslint rule disabled, we have to technical mechanism of that there are things still not final.

layouts/default.vue
SimonSimCity: Please revert these changes. They are reported as warning by eslint because we use `console`, but its not an error and this shouldn't stop you from contributing.

@sifferhans Related to https://github.com/bcc-code/bmm-web/issues/33 - we should define a strategy of how to deal with errors of various kinds.

SimonSimCity (changes requested):
Just these three changes and I'm fine:
- eslint rule
- `light` tag
- error-handling of `NuxtErrorBoundary`

layouts/default.vue
sifferhans: Nuxt has a `NuxtErrorBoundary` component that catches errors from child components, which has a `@error` event. We could look into using this for client side errors, but we probably need a separate strategy for API errors

SimonSimCity: @sifferhans See https://github.com/bcc-code/bmm-web/issues/33#issuecomment-1531143942. We should move this discussion, its off topic here.

components/AppToolbar.vue
belljohan3: Ok. Which text should be translated please ?

package.json
```diff
@@ -19,8 +19,12 @@
   },
   "dependencies": {
     "@auth0/auth0-vue": "^2.2.0",
+<<<<<<< HEAD
```
SimonSimCity, line 22: Seems like you have not resolved all the merge-conflicts 😅

SimonSimCity (changes requested):
Would be nice to get this PR merged soon. Please take a look at the open points and discuss or change where necessary.

pnpm-lock.yaml
```diff
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
belljohan3: The conflict has been resolved now.

package.json
belljohan3: Yes, i have to resolve it.

components/ProfileMenu.vue
```diff
@@ -0,0 +1,119 @@
...
+        enter-to-class="transform scale-100 opacity-100"
+        leave-active-class="transition duration-75 ease-in"
+        leave-from-class="transform scale-100 opacity-100"
+        leave-to-class="transform scale-95 opacity-0"
+      >
+        <MenuItems
+          class="absolute right-0 mt-2 w-56 origin-top-right divide-y divide-gray-100 rounded-md bg-white shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none"
+        >
+          <div class="px-1 py-1">
+            <MenuItem v-slot="{ active }">
```
SimonSimCity, line 29: All those menu items seem to have the same property. Can't we use a loop here..?

I'd only include options which are done yet or show some kind of message - not something which we are about to implement.

SimonSimCity (changes requested):

components/ProfileMenu.vue
belljohan3: I just look at the headlessui documentation. I would like also to refact it to have less code. Would you let me know how to do it ? 
-------------------
Also, i've installed nuxt-headlessui module and set it up in nuxt.config.ts file. But i am still getting red line when i am trying to import elements i need for the menu

PR #104, opened 2023-05-02
Add queue support to mediaPlayer

u12206050:
Add shuffle to easy test queue
Add next and previous to composable, not sure if we have a need for it yet.


SimonSimCity:
LGTM as a first draft. I'd further on try decoupling the queue from the player while keeping them as a single interface on the plugin. This way we can have an easier time testing while maintaining an easy 2 use interface.

Please let prettier and eslint do their job and I'm fine with merging it in.

@kkuepper maybe good to streamline the source logic with what's in the app already, so the UX is equal?

u12206050:
@SimonSimCity Havn't done anything to hinder prettier and linting, although it doesn't seem to work when I run `pnpm lint` is there something in the install or script missing, alternatively some missing steps in README. Ideally anyone starting off with the project should have everything working out of the box when following README steps.
Maybe we can take a look at it tonight.

u12206050:
btw, what is up with all the weird file names `3.mediaPlayer` etc?

SimonSimCity:
> it doesn't seem to work when I run `pnpm lint` [...]

Well, that's for checking. All linting should work out of the box when using vscode and the recommended extensions. If you disabled auto-run of prettier when saving, you'll (obviously) have to run it on your own, which seems to be the case in your env.

Have to check if I can make it to WWW tonight.

> btw, what is up with all the weird file names `3.mediaPlayer` etc?

If you mean the prefix of numbers, that's the nuxt-way to ensure plugins are loaded after those they depend on: https://nuxt.com/docs/guide/directory-structure/plugins#plugin-registration-order

plugins/3.mediaPlayer.ts
```diff
@@ -44,65 +46,119 @@ export default defineNuxtPlugin((nuxtApp) => {
...
+    // else move track to end of queue only if it is before the current track
+    const index = queue.value.findIndex((t) => t.id === track.id);
+    if (index === -1) {
+      queue.value.push(track);
+    } else if (index < currentQueueIndex.value) {
+      queue.value.splice(index, 1);
+      queue.value.push(track);
+    }
+
+    // Play if currently paused and no track is playing
```
SimonSimCity, line 142: I guess I'll have to rip this apart in another PR... I can just come up with all to many scenarios where this wouldn't be what a user would expect to happen... E.g. when playing some songs but clicking on a song of an album ...

SimonSimCity (changes requested):
Looking through the action-flow, this looks rather superficially implemented, just supporting one use-case 🧐 We can merge this, but I'll take this apart later either way to support more use-cases, which might require a full refactoring of the code merged in here...

pages/playlist/curated/[id].vue
```diff
@@ -8,6 +10,18 @@ const playlistId = Number(id);
...
 
+const { setCurrentTrack, addTrackToQueue } = inject(MediaPlaylistInjectionKey)!;
+
+function shuffle() {
+  // shuffle tracks and add them to the queue
+  const shuffledTracks = tracks.value?.filter((track) => true);
+  if (shuffledTracks && shuffledTracks.length > 0) {
+    shuffledTracks?.sort(() => Math.random() - 0.5);
+    setCurrentTrack(shuffledTracks.pop()!);
+    shuffledTracks.forEach((track) => addTrackToQueue(track));
```
Comment (kkuepper) [inline pages/playlist/curated/[id].vue:21]: I think we need a way to add multiple tracks to the queue at once.
Something like:
playTracks(Track currentTrack, List<Track> queue)

Comment (kkuepper) [reply pages/playlist/curated/[id].vue:21]: But I think this PR did what it wanted to achieve and further steps can be done in a separate PR.

kkuepper:
@SimonSimCity I would merge this and do further changes in a separate PR. Is that ok?

SimonSimCity:
@kkuepper I agree after resolving my question about the changes in the `.gitignore` file. Those changes, to me, are a sign of that the environment is not set up for DX with nuxt but for speeding up things which can not properly be done in a nuxt-env.

kkuepper:
reverted .gitignore and made pnpm lint pass.
@SimonSimCity If you or someone else doesn't complain the next few hours, I'll merge this PR.

SimonSimCity (approved):

SimonSimCity:
Merging it. https://github.com/bcc-code/bmm-web/pull/104#pullrequestreview-1410900159 should be picked up in a separate PR then.

PR #114, opened 2023-05-05
bugfix/track dropdown menu

sifferhans:
- clean up template
- implement design like in figma

image

### DIFF
--- a/components/DropdownMenu.vue
+++ b/components/DropdownMenu.vue
@@ -19,6 +19,7 @@ function close(event: Event) {
   emit("close");
 }
 
+// menu items
 function menuItemClick(event: Event, item: DropdownMenuItem) {
   close(event);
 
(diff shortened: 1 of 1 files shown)


SimonSimCity (approved):

SimonSimCity:
Just since you're already on refactoring things ...

Can we please agree on a strategy to have this more visible if a function stops the propagation or not? As of now, I'd have to look into `menuItemClick` which calls `close` to validate that propagations are also stopped at that branch... That's tedious.

sifferhans:
@SimonSimCity We should probably look into using Headless UI or something like that (#101 has started with something like this) for the dropdown.

We should probably create a standardised dropdown component so we stay consistent with styling and functionality

kkuepper (approved):

PR #116, opened 2023-05-10
Update sdk

SimonSimCity:
Also updated async-data calls to be lazy. The lazy version doesn't block the transition of routing, which makes for a better user-experience if the user does not have to remain on the old view until all data for the new route is loaded.

### DIFF
--- a/components/ProtectedImage.vue
+++ b/components/ProtectedImage.vue
@@ -7,7 +7,7 @@ const props = defineProps<{
   alt?: string;
 }>();
 
-const source = ref<string>();
+const source = ref<string>("");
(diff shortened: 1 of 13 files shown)


kkuepper (approved):
This is awesome! I tried to achieve the same but you managed it much more efficient.
Can we merge this right away?

SimonSimCity:
Most of my changes you could even be taking into the API repository, what do you think @kkuepper ?

kkuepper:
I guess you mean the changes to the swagger?
In theory yes. But in practice it's hard to know what exactly needs to change to generate the right swagger. If you want to give it a go, you can create a PR.

PR #120, opened 2023-05-22
Profile dropdown component

belljohan3:
Hi @SimonSimCity ! 
cc: @kkuepper @sifferhans  @andreasgangso 
I've some issues. I readed at the documentation of Headlessui and i implemented it in the code as peer as the doc. But, sometimes, when i click on `Profile` the menu is dropping down and sometime it doesn't. I don't really what the problem is but i can't move forward because i can't see what i am doing.


SimonSimCity:
When running your code, it doesn't open at all, but every time I click on it, I can see the following error message in the console:
Screenshot 2023-05-23 at 15 21 14

As I've tested it, the library seems to have a problem rendering multiple nodes within a node defined using `as="template"` which by default applies to the components `Menu` (where you changed it to `div`) and `MenuItem` (where it crashes right now). I've already reported this bug to the developers of the library: https://github.com/tailwindlabs/headlessui/issues/2512

Is this a limitation we can live with for the time being?

sifferhans:
@SimonSimCity I think that's a limitation of the library, but it shouldn't be a problem. 
Just set `as="li"` prop on the `MenuItem` that has multiple children

belljohan3:
@sifferhans Does it mean that if i change `as=div` to `as=li` it will fix the problem ? Also, the Switch component component don't render in the browser. May be it's due of the same reason ?

sifferhans:
@belljohan3 If you look at the error message Simon posted earlier it actually explains the problem quite well.
You need to add an `as` prop to the `MenuItem` components with multiple children (probably only the one with a switch)

belljohan3:
@sifferhans Okay I see. I will work on it tonight.🎉

belljohan3:
@sifferhans i have added the prop `as=li` in the `MenuItems` component but it doesn't work. I think it is a component that have many children. Did I put in the wrong tag ?

sifferhans:
> @sifferhans i have added the prop `as=li` in the `MenuItems` component but it doesn't work. I think it is a component that have many children. Did I put in the wrong tag ?

Yeah, it should be on each `MenuItem` component, not the `MenuItems` component

sifferhans:
Thats right 😁

SimonSimCity:
Please be careful when merging! The last commit to me looks like an attempt of merging, where you accidentally removed some of your changes and changed the commit message, maybe more 😅🙈

Tip: use a 3-way-diff for resolving merge conflicts when things get too messy 🤗

belljohan3:
I used that way to resolve conflict. I don't know what happen exactly.

SimonSimCity:
Theses changes aim for fixing #8. All not solved tasks should be noted down and taken up later. Here's a small list:

- [ ] Implement switch for dark-mode, replacing the switch introduced by #152
- [ ] Implement switch for app-language, replacing the switch introduced by #51
- [ ] Implement switch for content-languages and adjust settings of API (see plugin `2.bmm-api`)
- [ ] Implement `Feedback` @sclausendk - needs clarification
- [ ] Implement `Contact` @sclausendk - needs clarification
- [ ] Implement `About BMM` @sclausendk - needs clarification

SimonSimCity (approved):

kkuepper (commented):
- Contact should have mailto:bmm-support@bcc.no link
- --Feedback-- should be 'User Voice' with a link to https://uservoice.bcc.no/?tags=bmm
- Remove 'About BMM' for now. My guess is that we'll replace it with a copyright page. But we don't need to worry about that until v0.6

SimonSimCity:
Just for reference because I was looking for why one would add `target="_blank"` to a `mailto`-link - https://stackoverflow.com/a/9880404/517914 - the comment is hilarious 😂

Btw, on MacOS a new tab was not opened on Firefox and Safari. Chrome opened a new tab but immediately closed it after recognising the protocol was to be handled outside the browser.

If the `About BMM` should contain a copyright, we can (and should) add a ticket for it, can't we ..?

kkuepper (approved):

kkuepper (approved):
looks good.

I moved leftover tasks into #162

PR #122, opened 2023-05-24
Fix electron login on built version and other small adjustments...

### LINKED ISSUE #71: setup electron

SimonSimCity:
* Use electrons error-message if there's an error
* Fix login redirection
* Add app-icon

Fixes #71


SimonSimCity:
Maybe a good option to include the build also in the checks ... it shouldn't fail, but it might 😅 It would also always give us an artifact to work with 😏

SimonSimCity:
After merging all these changes, everyone having write-access should be able to manually run the workflow `Build/release app` from within the `Actions` panel.

kkuepper (approved):

kkuepper:
@steffanhalv Do you by chance know how to force a new login? Reinstalling didn't ask me to login again making it annoying to work with the login.

steffanhalv:
I think we probably should add a logout button

SimonSimCity:
@kkuepper We could just implement the `logout` function of the auth0 plugin 😏

PR #126, opened 2023-05-26
create signed dmg for Mac

kkuepper:
creates a signed .dmg for arm and Intel Macs.
The windows .msi still has problems during the login process but on Mac it should work fine.

Have a look at https://github.com/bcc-code/bmm-web/releases for the output.


kkuepper:
Maybe we should squash this PR to hide all the trial and error commits.

.github/workflows/electron.yml
```diff
@@ -29,14 +29,34 @@ jobs:
...
+        env:
+          CSC_LINK: developer-id.privatekey.p12
+          CSC_KEY_PASSWORD: ${{secrets.APPLE_DEVELOPER_ID_PASSWORD}}
+          APPLE_ID: ${{secrets.APPLE_ID}}
+          APPLE_ID_PASSWORD: ${{secrets.APPLE_ID_PASSWORD}}
+          APPLE_TEAM_ID: KJ6LCYQ3A8
+
+      - name: Building the Electron app for Windows
         run: npx electron-builder
+        if: matrix.os == 'windows-latest'
```
SimonSimCity, line 55: Please try adding back Ubuntu and change this condition to run for both.. it does doesn't really cost us anything to have it, does it..?

SimonSimCity (commented):
This doesn't auto-publish but just create signed artifacts, right?

.github/workflows/electron.yml
kkuepper: Doesn't cost anything. But do we have anybody that can verify if it's working on Linux?
If not we can't publish it anyway.

kkuepper:
> This doesn't auto-publish but just create signed artifacts, right?

I believe for builds from the main branch or with the right tag are published by default.
But I mainly focused on getting it signed for this PR.

.github/workflows/electron.yml
SimonSimCity: @Ghostbird could install the version I created earlier, so I don't know if there's anything additionally needed. Seems to be just fine using the same commands as for windows (which didn't need any changes, right?)

SimonSimCity:
I've just added the handling of the login - now it works for Mac in DEV again 😎 To be tested in other environments.

Just had it sometimes that the app didn't load properly (remained in loading-stage) or redirected the user twice to the browser and back to the app. This needs some further investigation, but the main process seems to work.

PR #127, opened 2023-05-26
Electron - login in browser

### LINKED ISSUE #124: Electron - use system default browser for logging in
Of the social logins, at least google seems to have problems with our electron setup.

We can use the default system browser for logging in (where the user might already be logged in) which will also have the users session on social logins already in place. This might help setting it up: https://stackoverflow.com/questions/31749625/make-a-link-from-electron-open-in-browser - and on going back to the custom protocol, it will open the app again.

SimonSimCity:
Fixes #124 and closes a gap between dev and prod. Both now use the custom protocol for reaching out for internal data.


steffanhalv (approved):
Nice, looks good to me 👍

Issue #64, opened 2023-04-21
load home from server


kkuepper:
- bmm-api.brunstad.org/Discover
- age can be ignored for now
- order of components and content is taken from /Disover

kkuepper:
currently blocked because of the swagger

SimonSimCity:
Can be worked on after merging PR #116

SimonSimCity:
Fixed by #129, future things have been separated in their own tickets.

PR #129, opened 2023-05-30
Feature/home layout loading

SimonSimCity:
Solves #64 to quite an extend - not fully yet ... someone with experience in tailwind should probably have a closer look at my work and do some follow-up changes here - I just can't get comfortable with it ... 😅 

What's still open to fully solve #64:

- [ ] Find out how to get the age of a person
- [x] Reload when locale changes
- [x] Fix styling for `album` and `playlist`
- [x] Confirm design for `track`
- [x] Implement design, route and component for `contributor` and `podcasts` (should go into a separate issue; both are very similar to a curated playlist) - #132, #133

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -2,56 +2,18 @@
 const { t } = useI18n();
 toolbarTitleStore().setReactiveToolbarTitle(() => t("nav.home"));
 
-// Playlists
-const { data: playlists, pending: playlistsPending } = useCuratedPlaylists();
-const newestPlaylists = computed(() => playlists.value?.splice(0, 4) || null);
-
-// Speeches
-const { data: speeches, pending: speechesPending } = useTracks({
-  contentType2: ["speech"],
-  size: 5,
-});
-
-// Audiobooks
-const { data: audiobooks, pending: audiobooksPending } = useTracks({
-  contentType2: ["audiobook"],
-  size: 5,
-});
+const { data: discoveries } = useDiscover();
 </script>
 
 <template>
   <div class="flex flex-col gap-16">
-    <ContentSection title="Playlists">
-      <div v-if="playlistsPending" class="flex h-64 gap-6">
-        <div
-          v-for="i in 4"
-          :key="i"
-          class="aspect-square h-full rounded-xl bg-background-2"
-        ></div>
-      </div>
-      <PlaylistCarousel
-        v-else-if="newestPlaylists"
-        :playlists="newestPlaylists"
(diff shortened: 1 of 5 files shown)


sifferhans (approved):
Looks good to me. 
Nice improvement to i18n too :)

SimonSimCity:
> Nice improvement to i18n too :) Should probably be a separate PR if we were to be that strict, but no matter

IMO not, because it was needed for this PR: https://github.com/bcc-code/bmm-web/pull/129/files#diff-02281a80fab758cb4e8e8359b222a8a5afeaa9d5a7ba858f15d852f1f8969ecfR9

pages/index.vue
```diff
@@ -1,57 +1,104 @@
...
+            <PageHeading v-if="item.type === 'section_header'" :level="3"
+              ><a v-if="typeof item.link === 'string'" :href="item.link"
+                >{{ item.title }}
+              </a>
+              <span v-else>{{ item.title }}</span></PageHeading
+            >
+            <NuxtLink
+              v-else-if="item.type === 'album'"
+              :to="{ name: 'album-id', params: { id: item.id } }"
+              style="display: inline-block; padding: 0 20px 20px 0"
```
sifferhans, line 67: Here you should use `class="inline-block pr-4 pb-4"` instead of inline styles

pages/index.vue
```diff
@@ -1,57 +1,104 @@
...
+            >
+              <ProtectedImage
+                :src="item.cover || ''"
+                style="
+                  max-width: 214px;
+                  border: 1px solid rgba(129, 136, 143, 0.1);
+                  border-radius: 16px;
```
sifferhans, line 74: Probably something like `class="max-w-[214px] border rounded-md"`

pages/index.vue
SimonSimCity: This really looks like writing styles in the css property 🤪😂

pages/index.vue
SimonSimCity: Why use `rounded-md` here? Isn't the `-md` equal to `0.375rem`, which is `6px` for most systems..? And adding `border` here gives it a solid black border, not the color I got in the design.

SimonSimCity:
I've changed quite some stuff for reusability, added grouping and also added a well visible note `Not yet implemented` for Podcasts (even though its visibly correct) and non-implmeneted types as contributors.

sifferhans (approved):

PR #131, opened 2023-05-31
Better ci error handling

SimonSimCity:


SimonSimCity:
What's daunting now is that the job `Deploy to Azure Static Web Apps` is marked successful ... 😔

PR #140, opened 2023-06-12
Tauri, just a demo, don't merge

kklem0:
It creates a 5 MB desktop app with 30 MB memory usage, also mobile if you use tauri beta.


SimonSimCity:
Could you add or link some resources of how this could be tested? We're well on our way of integrating electron, but if Tauri is well enough and we can integrate and build it, why not spend more time looking into it and drop electron ..?

SimonSimCity:
[Seems Tauri doesn't seem to support deep-links ...](https://github.com/tauri-apps/tauri/issues/323) Does this mean, it would not be easy to have the login process in the browser?

kklem0:
> [Seems Tauri doesn't seem to support deep-links ...](https://github.com/tauri-apps/tauri/issues/323) Does this mean, it would not be easy to have the login process in the browser?

So you need to use https://github.com/FabianLars/tauri-plugin-deep-link and to intercept external traffic similar to Electron.
But I'm wondering if we could implement the login process same as BANK ID in Norway which is using an iframe instead of browser.

Screenshot 2023-06-15 at 14 55 58

kklem0:
> Could you add or link some resources of how this could be tested? We're well on our way of integrating electron, but if Tauri is well enough and we can integrate and build it, why not spend more time looking into it and drop electron ..?

To me it never hurts to try new tech and be open to new things, a few new things that I'm using these days started using Tauri and I can see the much smaller package size with decent performance, so it's nice to have a project to test it out. It works with bmm-web for now but not auth, I injected the auth token in to make it work, but I can see other projects made it work with 3rd party plugin with deep linking so it's nice to test it out.

SimonSimCity:
Well, we had it earlier that the login process is fully handled by the electron app, but we had troubles with the google login. Since credentials are anyways usually easily accessible in the browser, we chose to take the step and integrated the system default browser into the login process. You could also simply have it different. There's also other methods for logging in (see `/app.vue` which has the method `loginWithRedirect()`) ... Having it in the browser just feels that the user can use all completion-additions and password-managers he's already so used to 🤗

kkuepper:
Nothing has happened for several months.

PR #142, opened 2023-06-14
show correct track duration & show album name in some cases

kkuepper:


components/track/TrackList.vue
```diff
@@ -13,9 +13,22 @@ withDefaults(
     skeletonCount: 5,
   }
 );
+
 const { setCurrentTrack, addTrackToQueue } = inject(MediaPlaylistInjectionKey)!;
 const showDropDownForTrack: Ref<null | string> = ref(null);
 
+const isTrackTypeKnown = () => {
+  const firstType = props.tracks?.[0]?.subtype;
+  console.log("first type", firstType);
```
SimonSimCity, line 22: To me this doesn't look like its very valuable if this function is working as expected. Please remove before merging or make it more intuitive to read and relate.

components/track/TrackList.vue
```diff
@@ -13,9 +13,22 @@ withDefaults(
...
+  const tracksWithDifferentType = props.tracks?.filter(
```
SimonSimCity, line 23: Just a minor note: If you're not interested in the actual amount, why not use [`find()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) or even better: [`every()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) or [`some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)? Any of them would stop after the first occurrence - the latter make the code even more intuitive to read IMO.

Maybe selecting a matcher-function based on the first type makes it even better to read - at least quicker to run, but I'll leave this over to you 😄

sifferhans (approved):
Looks good:)

components/track/TrackList.vue
kkuepper: I guess you mean the console.log that I overlooked.

components/track/TrackList.vue
kkuepper: I didn't remember what functions are available and what they are called. Thanks for the hint.

SimonSimCity (approved):

SimonSimCity:
LGTM - just one last question: What are the changes in the `pnpm-lock.yaml` file for if the `package.json` didn't change ..? Is this a left-over?

kkuepper:
Yes. That probably should have been part of the previous PR.

Issue #128, opened 2023-05-30
Electron - show info to user before redirecting to browser for login


SimonSimCity:
The login process works, but might be not as intuitive to a normal user as it's now. The goal of this ticket is to show an information to the user telling what is about to happen before navigating from the app to the browser and back.

https://github.com/bcc-code/bmm-web/commit/c416137d03bfcaff5b473f0479ce55f6fd91db58#diff-13ee644059574b054a4e5ec081666a03579b28da03a50331215908f8d0ef7cfdR35-R36

SimonSimCity:
We'll just open the browser and leave a message for the user in the electron window of that the login is in progress and how he can reinitiate the process if he accidentally closed the browser window.

PR #143, opened 2023-06-15
Handling of links and login in electron

### LINKED ISSUE #123: Deep links (electron)
Electron allows us to use deep links. Different from a mobile app, it seems that we have to use custom protocols for this. A good write-up can be found here: https://shipshape.io/blog/launch-electron-app-from-browser-custom-protocol/

SimonSimCity:
Did quite some changes on handling links which sum up in that links from external sources (another application opens `bmm://` links) and external links (links to `bmm.brunstad.org` via `http` or `https`) are handled by vues router.

The message shown to the user in electron while the login is in progress could do well with some styling, but I'll for now leave it up to @sclausendk to define how it should look like.

I've also added a long awaited logout link - just for having one...

Fixes #123 and #128

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -25,7 +25,7 @@ useHead({
       <NuxtPage />
     </NuxtLayout>
     <div v-if="!isAuthenticated">
-      You are redirected to the login page. Please wait or
+      You will be redirected to the login page. Please wait or
       <a href="#" @click="loginWithRedirect()">try again</a>.
     </div>
   </div>
--- a/app.vue
+++ b/app.vue
@@ -20,11 +20,15 @@ useHead({
 </script>
 
 <template>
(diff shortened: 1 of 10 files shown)


app.vue
```diff
@@ -20,11 +20,15 @@ useHead({
...
 
 <template>
-  <NuxtLayout v-if="isAuthenticated">
-    <div class="container mx-auto p-2 lg:p-5">
+  <div class="container mx-auto p-2 lg:p-5">
+    <NuxtLayout v-if="isAuthenticated">
       <NuxtPage />
+    </NuxtLayout>
+    <div v-if="!isAuthenticated">
+      You are redirected to the login page. Please wait or
```
kkuepper, line 28: You will be redirected ...

kkuepper (approved):

PR #144, opened 2023-06-15
Flexible version in sidebar

SimonSimCity:
The version number is now taken from the `package.json` file. Reduces the number of places we have to adjust when publishing.


kkuepper (approved):
I was just about to start the same PR

PR #148, opened 2023-06-16
Windows deep links

SimonSimCity:
Tested on MacOS (Intel) and Windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ npx electron-builder
 
 _Note:_ The electron-builder will by default only build an app for the current platform and current architecture. If you want to go beyond, please visit https://www.electron.build/multi-platform-build.html.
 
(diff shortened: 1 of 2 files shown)


SimonSimCity:
It seems the protocol is not registered during installation of the app (https://stackoverflow.com/questions/45809064/registering-custom-protocol-at-installation-process-in-electron-app) - which opens another question for me:

Is the protocol only registered while running the app, or is it registered on first start and also usable when the app is not running?

To answer this question, I need an installable build.

SimonSimCity:
On MacOS and Windows the protocol is registered after opening it for the first time.

With the latest changes the app doesn't crash anymore on MacOS if the last window is closed; opening deep-links without having the app running works also as expected, so MacOS is working 😎🎉

Windows now also works mostly. The only downside is that the app starts with the main page when opened by a deep-link. It opens the correct page when opening a deep-link while the app is running though... Still needs some fine-tuning here...

kkuepper (commented):
Now everything is in main.ts. It would be nice to extract the part with deeplinking into it's own file.

SimonSimCity:
> Now everything is in `main.ts`. It would be nice to extract the part with deep-linking into it's own file.

Well ... that's almost be the whole file ... Feel free to split it up as you see it fit though 🤗

kkuepper (approved):

PR #150, opened 2023-06-22
Appimage package improvements

Ghostbird:
Minor improvements improve how the produced AppImage integrates with most XDG compatible desktops.

Didn't manage to fully fix URI handling, but @SimonSimCity's branch has pretty much the same code that I wrote, so I threw mine out.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -23,5 +23,6 @@ linux:
   desktop:
     Keywords: audio;bcc;bmm;brunstad;christian;church;edification;faith;media;music;sermon
     SingleMainWindow: true
+    StartupWMClass: bmm-web
   target:
     - AppImage
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -19,5 +19,8 @@ win:
     - msi
     - zip
 linux:
+  category: Audio;Player
+  desktop:
+    Keywords: audio;bcc;bmm;brunstad;christian;church;edification;faith;media;music;sermon
   target:
     - AppImage
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -22,5 +22,6 @@ linux:
   category: Audio;Player
   desktop:
     Keywords: audio;bcc;bmm;brunstad;christian;church;edification;faith;media;music;sermon
+    SingleMainWindow: true
   target:
     - AppImage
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
     "build": "nuxt generate",
     "dev:electron": "cross-env ELECTRON=true nuxt --port 9001 dev",
     "build:electron": "cross-env ELECTRON=true nuxt generate",
+    "package:electron": "node node_modules/electron-builder/cli.js",
```
SimonSimCity, line 13: I'd prefer using the command `electron-builder` (it's inside a npm command context, so no need to prefix with `npx`) here instead of referencing the script the package `electron-builder` defines as `bin`. Not much of a change though, but rather a thing of flexibility...

electron-builder.yml
```diff
@@ -19,5 +19,10 @@ win:
     - msi
     - zip
 linux:
+  category: Audio;Player
+  desktop:
+    Keywords: bcc;bmm;electron;media;audio;christian;faith;speech;music
```
SimonSimCity, line 24: I'd remove `electron` from this list 😅 Rest looks good to me 😎 Maybe `Brunstad` would be a good addition ...

SimonSimCity:
The e2e test is failing because env-variables are not available on foreign PRs and its failure should therefore not stop this PR from being merged.

electron-builder.yml
Ghostbird: I included `electron` because an example I saw did it too, but I can remove it.
The main use for these, is that your shell can match these when you search for applications:

image

> **Note**
> This example uses a different BMM.desktop file that I wrote myself to integrate an older release. The icon and text doesn't match the desktop file generated by this branch, but it demonstrates the point.

SimonSimCity: Its fine IMO. Just find it a bit daunting when apps appear in a search which seemingly have no relation to the keyword ... I wouldn't expect anything built with `java` to appear when I acutally just want to start something having `java` in its name, but that's me ... 🤗

package.json
Ghostbird: Ah, I didn't know that the npm context worked like that. If that works, it's much better.

electron-builder.yml
Ghostbird: Changed the keyword list to `audio;bcc;bmm;brunstad;christian;church;edification;faith;media;music;sermon`

- Removed: _electron_
- Added: _edification_
- Replaced: _speech_ which is more common for TTS / speech synthesis whit _sermon_
- Sorted alphabetically

package.json
Ghostbird: Yep, changed it, and it works nicely.

Issue #135, opened 2023-06-01
Send age of user to API on home-screen


SimonSimCity:
The API request for the discover-endpoint, used at the home-screen, has an option to hand in the users age. Since it's not required, it's also doable without. Since the app seems to feed the API with this info, it would be good for an equal user-experience, that this info is passed on to the API on the website as well.

https://github.com/bcc-code/bmm-web/pull/129/files#diff-02281a80fab758cb4e8e8359b222a8a5afeaa9d5a7ba858f15d852f1f8969ecfR16

PR #157, opened 2023-06-27
add analytics with ApplicationInsights

kkuepper:


plugins/4.applicationInsights.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    router.afterEach((route: { name: any }) => {
+      appInsights.stopTrackPage(route.name);
+    });
+  };
+
+  const appInsights = new ApplicationInsights({
+    config: {
+      connectionString:
+        "InstrumentationKey=45d71410-6840-472a-bf95-45f77749c0a2;IngestionEndpoint=https://westeurope-5.in.applicationinsights.azure.com/;LiveEndpoint=https://westeurope.livediagnostics.monitor.azure.com/",
```
sifferhans, line 72: This connection-string should perhaps be an environment variable?

plugins/4.applicationInsights.ts
```diff
@@ -0,0 +1,83 @@
...
+      /* ...Other Configuration Options... */
+    },
+  });
+  appInsights.loadAppInsights();
+  appInsights.trackPageView();
+
+  inject("appInsights", appInsights);
```
sifferhans, line 79: What is this `inject('appInsights', appInsights)` ?

sifferhans (approved):

plugins/4.applicationInsights.ts
```diff
@@ -0,0 +1,83 @@
+// This plugin was inspired by https://github.com/nuxt-community/applicationinsights-module
```
SimonSimCity, line 1: What's actually the reason for not using the original package ..?

kkuepper: It doesn't work with Nuxt 3

plugins/4.applicationInsights.ts
kkuepper: I assumed that it makes it usable in other components. But I haven't gotten that far yet.
There's a reason it's only a draft PR. I still need to figure out how to provide additional info and how to create custom events.

sifferhans: I see! You can return it like this from the plugin:

```ts
return {
	provide: {
		appInsights
	}
}
```

You will then have access to it through the `useNuxtApp()` composable:

```ts
const { $appInsights } = useNuxtApp()
```

kkuepper: What's the best way to use appInsights in the 3.mediaPlayer.ts plugin?

kkuepper:
@sifferhans How can I get strong typing for $appInsights? Or should I use a different mechanism?
Screenshot 2023-06-28 at 3 33 32 PM

SimonSimCity:
@kkuepper updating this branch with the latest TypeScript changes (#159) should do half the job. Another part is the file `.nuxt/components.plugin.mjs`, which I'm on... After adding `.nuxt/components.plugin.ts` with an empty object as export the completion worked for me, excepted for in vue-files 😅🙃 trying to fix that as well...

plugins/4.applicationInsights.ts
SimonSimCity: @kkuepper https://nuxt.com/docs/guide/directory-structure/plugins#typing-plugins - combined with the statement above the anchor, which is what Sigve was pointing out. Auto completion for those helpers is still tricky 🙈

Injecting it the way you did here (https://vuejs.org/guide/components/provide-inject.html) is the Vue-way, but Nuxt has different way, which is the one Sigve mentioned 🤗

SimonSimCity:
I've squashed the changes, moved the connection string to the environment variable `NUXT_PUBLIC_APPLICATION_INSIGHTS`, moved the extra dependency `nuxt` back as dev-dependency and fixed some additional imports.

SimonSimCity:
It just seems that despite of me adding it as variable in the settings of this project, it doesn't seem to be available in the process, because otherwise the tests of [5817d90](https://github.com/bcc-code/bmm-web/pull/157/commits/5817d90b0e4c83f5a874f8fde469ff3f72252d6d) should not have failed ...

SimonSimCity (approved):

SimonSimCity:
Fixes #135

PR #158, opened 2023-06-27
Package for Debian

Ghostbird:
When the deb package is installed, the Desktop file that contains the URI handler line is automatically installed in the system default location. Then if you have `desktop-file-utils` installed (pretty much default) this will kick off an APT trigger that updates the system's MIME associations. As a consequence, you _need_ `root` access on the machine to install the deb file, but it will result in a system-wide install that has the URI handler registered.

## Rationale

On Linux you could use the AppImage file. Provided you have first installed AppImageLauncher. AppImageLauncher currently has issues if newly installed on a Debian 12 "Bookworm" OS. Probably related to the latter's move from GTK3 to GTK4.

This bug won't affect anyone who:

- Already had AppImageLauncher correctly configured
- Can use all the default settings of AppImageLauncher

An attempt to configure AppImageLauncher on first start-up will crash. I had to manually write the necessary configuration files to be able to use AppImageLauncher to be able to install‌ BMM. It was, _less work_ to figure out how to configure `electron-builder` to build a correct Debian package next to the AppImage, than it was to install the AppImage on my system. This included two issues I had to solve first:

- Our electron builder was an older version that didn't support all the appropriate configuration
- Even the current electron-builder doesn't property handle building a deb package for an icon that's 1024×1024px.

I haven't been able to test whether the upgrade of `electron-builder` was _completely_ safe for other OS builds. None of the builds failed, and the archive contains the files, but I haven't tested those files.

I have managed to download the archive and install the `bmm-web_0.0.3_amd64.deb` on another machine than where I developed it, which worked perfectly.

@SimonSimCity Can you check on this?


SimonSimCity:
LGTM. I can just test this on other OS, but I don't see a reason to fail, so let's merge.

PR #160, opened 2023-06-28
Media keys

### LINKED ISSUE #136: Electron - Support for media keys
We have not yet spend any time to support media keys in the desktop-app. Here's some feedback I got from @Ghostbird (Linux):

>Integration with the OS media features is partial. Play/pause and stop button work. Next and previous track buttons don't.
Physical media key events sent from the keyboard and mouse don't work, not even play/pause.

A good outcome would be when the media-keys for Play/pause, stop, next and previous are working - whether the app is active or not - based on usage, we'll focus on Mac, Windows; Linux is optional.

SimonSimCity:
Fixes #136.

As far as I could test, registering the DOM Media Session already fixed all of our requests for handling media keys - at least on MacOS.

If needed, I'll later include the registering of the media-keys, but at least on MacOS none of them could be registered, and the DOM Media Session did the job quite well.
[electron-media-keys.patch](https://github.com/bcc-code/bmm-web/files/11892859/electron-media-keys.patch)

Because of some quirx in typescript, this PR includes #159, which should be merged first.

### DIFF
--- a/plugins/4.mediaKeys.ts
+++ b/plugins/4.mediaKeys.ts
@@ -1,0 +1,79 @@
+import { defineNuxtPlugin, useNuxtApp } from "#app";
+import { MediaPlayerStatus } from "./3.mediaPlayer";
+
+export default defineNuxtPlugin(() => {
+  const { $mediaPlayer, $mediaPlaylist } = useNuxtApp();
+
+  // Register DOM Media Session, if available
+  if ("mediaSession" in navigator) {
+    watch(
+      $mediaPlaylist.currentTrack,
+      (track) => {
+        if (!track) {
+          navigator.mediaSession.metadata = null;
+          return;
+        }
+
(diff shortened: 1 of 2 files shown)


plugins/4.mediaKeys.ts
```diff
@@ -0,0 +1,79 @@
...
+      $mediaPlayer.play();
+    });
+    navigator.mediaSession.setActionHandler("pause", () => {
+      $mediaPlayer.pause();
+    });
+    navigator.mediaSession.setActionHandler("stop", () => {
+      $mediaPlayer.pause();
+    });
+    navigator.mediaSession.setActionHandler("seekbackward", (_) => {
+      console.error("The event `seekbackward` is implemented yet");
```
kkuepper, line 64: 'not' missing in error

kkuepper (approved):

Issue #115, opened 2023-05-10
Persistent storage for pinia


SimonSimCity:
We decided to store data in pinia. By default it is a non-persistent storage. To have the data persist, we can use an extension on top of it, like https://prazdevs.github.io/pinia-plugin-persistedstate/.

This extension needs to be configured and validated to work at least in a browser and in Electron.

sifferhans:
What are we to use Pinia for btw?

PR #164, opened 2023-07-01
Improved profile menu

sifferhans:
WIP


SimonSimCity:
Umm ... Why not use pinia and #115?

sifferhans:
@SimonSimCity Sure, could be better

sifferhans:
@SimonSimCity I use `@pinia-plugin-persistedstate/nuxt` to persist the autoplay setting now:)

locales/nb.json
SimonSimCity, line 1: @kkuepper can you comment on what effect this will have? Will this be taken in at crowding or is it going to be rejected? I feel it's good to know so we are aware of how the tools we have are aught to be used...

components/ProfileMenu.vue
```diff
@@ -34,64 +40,102 @@ const logout = async () => {
...
             <MenuItem v-slot="{ active }" as="div">
               <button
-                :class="active ? 'bg-violet-500 text-white' : 'text-gray-900'"
-                class="w-full px-2 py-2 text-left hover:bg-[red]"
-                @click.stop
+                :class="{
+                  'bg-label-separator dark:bg-label-dark-separator': active,
+                }"
+                class="w-full rounded-lg px-3 py-2 text-left"
+                @click="(e: MouseEvent) => {
```
SimonSimCity, line 120: Why was `@click.stop` replaced by this function?

app.vue
```diff
@@ -20,7 +20,7 @@ useHead({
 </script>
 
 <template>
-  <div>
+  <div class="h-full">
```
SimonSimCity, line 23: 👌

locales/nb.json
kkuepper: English is set as the source language. That means only changes to en.json will be used in Crowdin and the changes in this file will be overwritten / lost.

components/ProfileMenu.vue
sifferhans: Good question 🤔

SimonSimCity:
I just saw that some of the changes put in here are not related to the profile menu anymore... Since this is not the only open PR, please keep in mind that continuing this way will make you more likely to run into merge conflicts. Please stay focused and get this to a good merge-able state. Other improvements should go into separate PRs IMO.

sifferhans:
@SimonSimCity Which things should be removed from the PR?

SimonSimCity:
@sifferhans I'm specifically referring to the changes in `plugins/1.auth0.server.ts` and `components/album/SubAlbum.vue`. I guess both were done in an attempt to solve some formatting or type related things, but since the `lint` command on master goes through without any problem, I assume the changes here are not serving the requested purpose and are therefore not needed.

I've also seen you've introduced a new npm command `format`... What's the purpose this tries to solve? Isn't what you wanted already included in the command `lint:fix`?

SimonSimCity:
I guess here's some confusion here about the process: This might be different for other languages, but this yields for most typescript projects: Types are only part of the compilation and have no implication at runtime, because typescript is always compiled to JavaScript, which has no types. Formatting (prettier) and linting (eslint) doesn't have an implication either, as they are only used against typescript. After the compilation to JavaScript the code can be run in a runtime like node, or the browser (e.g. as e2e test or as in our live env).

This means, the e2e tests cannot be failing because of types, neither because of formatting or linting, but when the code does not what it's expected to do.

Btw, you can also download a video of the e2e tests to see where they fail. I guess it's still because BCC seemed to have changed something in the process of logging in; at least that was the case when I looked at it lately, and I just didn't have the time to fix it since, so that should be overlooked for now.

I'll have to check with BCC whether there's a chance to bypass this system, but I at least hope this process of logging in won't change that frequently.

sifferhans:
@SimonSimCity Yeah, it seems like I updated the lockfile by accident. My bad 😅

I don't really know if it is the project or if it is my VSCode setup that makes formatting so problematic, but I would guess it is my setup that is the culprit, as it seems like it only is a problem for me. I just cannot find what causes it 😆

sifferhans:
@SimonSimCity How do I undo the lockfile changes?

sifferhans:
@SimonSimCity Seems like only the e2e login part is failing now

components/ProfileMenu.vue
```diff
@@ -34,64 +40,94 @@ const logout = async () => {
...
+            <MenuItem v-slot="{ active }">
               <button
-                :class="active ? 'bg-violet-500 text-white' : 'text-gray-900'"
-                class="w-full px-2 py-2 text-left hover:bg-[red]"
-                @click.stop="autoPlayEnabled = !autoPlayEnabled"
+                :class="{
+                  'bg-label-separator dark:bg-label-dark-separator': active,
+                }"
+                class="flex w-full items-center justify-between rounded-lg px-3 py-2"
+                @click="
```
SimonSimCity, line 52: Why was `@click.stop` replaced by this function here as well ..?

components/ProfileMenu.vue
```diff
@@ -34,64 +40,94 @@ const logout = async () => {
...
+                :class="{
+                  'bg-label-separator dark:bg-label-dark-separator': active,
+                }"
+                class="w-full rounded-lg px-3 py-2 text-left"
                 @click.stop
               >
-                {{ $t("profile.theme") }}
+                <p>{{ $t("profile.theme") }}</p>
+                <span class="text-label-2 dark:text-label-dark-2">
+                  {{ $t(`profile.theme-${$colorMode.value}`) }}
```
SimonSimCity, line 91: Please do not use variables for translation text, otherwise our mechanism for detecting unused or missing language snippets (https://eslint-plugin-vue-i18n.intlify.dev/rules/no-unused-keys.html / https://eslint-plugin-vue-i18n.intlify.dev/rules/no-missing-keys.html) won't work.

components/ProfileMenu.vue
```diff
@@ -34,64 +40,94 @@ const logout = async () => {
...
+                :class="{
+                  'bg-label-separator dark:bg-label-dark-separator': active,
+                }"
+                class="w-full rounded-lg px-3 py-2 text-left"
                 @click.stop
               >
-                {{ $t("profile.app-language") }}
+                <p>{{ $t("profile.app-language") }}</p>
+                <span class="text-label-2 dark:text-label-dark-2">{{
+                  $i18n.locale
```
SimonSimCity, line 105: Should we just show the language code, or the actual name of the language?

components/ProfileMenu.vue
```diff
@@ -34,64 +40,94 @@ const logout = async () => {
...
+                :class="{
+                  'bg-label-separator dark:bg-label-dark-separator': active,
+                }"
+                class="w-full rounded-lg px-3 py-2 text-left"
                 @click.stop
               >
-                {{ $t("profile.content-language") }}
+                <p>{{ $t("profile.content-language") }}</p>
+                <span class="text-label-2 dark:text-label-dark-2">{{
+                  $i18n.locale
```
SimonSimCity, line 119: The value displayed here has nothing to do with the content-language. Please fix by either removing this value or replacing it with the correct value.

SimonSimCity (changes requested):

SimonSimCity:
Finally found some time to fix the e2e thingy and clean up the pnpm lock file - it's just hard to do it right ... but while re-reading the PR, I found some other things, which I'd like to see addressed before merging.

sifferhans:
@SimonSimCity Great, thank you 😁 I'll fix asap

components/ProfileMenu.vue
sifferhans: I think I display the correct values now, but the switching functionality should be implemented in a separate PR

components/ProfileMenu.vue
sifferhans: I've gone with `Intl.DisplayNames` for now, but we should probably get back to this later

components/ProfileMenu.vue
sifferhans: Changed to use a ternary instead of dynamic keys; is that ok enough?

components/ProfileMenu.vue
```diff
@@ -34,64 +44,98 @@ const logout = async () => {
...
+                :class="{
+                  'bg-label-separator dark:bg-label-dark-separator': active,
+                }"
+                class="w-full rounded-lg px-3 py-2 text-left"
                 @click.stop
               >
-                {{ $t("profile.content-language") }}
+                <p>{{ $t("profile.content-language") }}</p>
+                <span class="text-label-2 dark:text-label-dark-2">
+                  {{ contentLanguages.join(", ") }}
```
SimonSimCity, line 127: I know locales and translation is hard to get right 😅 Please read https://stackoverflow.com/questions/39237769/rtl-languages-comma-separated-lists and use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat or similar.

Not required for this being approved by me though, just a last touch to get it well 🤗

components/ProfileMenu.vue
```diff
@@ -34,64 +44,98 @@ const logout = async () => {
...
+                  'bg-label-separator dark:bg-label-dark-separator': active,
+                }"
+                class="w-full rounded-lg px-3 py-2 text-left"
                 @click.stop
               >
-                {{ $t("profile.theme") }}
+                <p>{{ $t("profile.theme") }}</p>
+                <span class="text-label-2 dark:text-label-dark-2">
+                  {{
+                    $colorMode.value === "light"
```
SimonSimCity, line 96: Well, `$colorMode.value` will give you the applied value, which doesn't reflect what this setting is about. As a user I'd like to see `system` here, when I've selected `system`, and not `dark`, just because my system is currently in dark mode. See https://color-mode.nuxtjs.org/#usage.

SimonSimCity (changes requested):
Almost there 😃🎉 Hope you don't feel all to stressed by all my requests for change 😅

sifferhans:
@SimonSimCity The color modes are now shown correctly (`system` is also shown)

SimonSimCity (approved):

PR #165, opened 2023-07-03
Package updates, electron config updates (v24)

SimonSimCity:
Not easy to see in the summary: I've updated many packages and also removed the package `electron-builder-notarize`, because `electron-builder` (since v24) already ships its own notarization tool - `@electron/notarize`

Pinning typescript version to 5.x fixes https://github.com/vuejs/language-tools/issues/2190


nuxt.config.ts
```diff
@@ -56,4 +56,12 @@ export default defineNuxtConfig({
   colorMode: {
     classSuffix: "",
   },
+  // TODO: Option is marked as invalid if env ELECTRON is not enabled, because this enables the module which allows the configuration.
```
kkuepper, line 59: not sure if I understand this

kkuepper (approved):

nuxt.config.ts
SimonSimCity: Nuxt generates types, specially those for the valid nuxt-config. If the module `nuxt-electron` is loaded, the config property `electron` is allowed. If its not added (when the env `ELECTRON` is falsy), this property is not valid. Means when running `pnpm dev`, your ide will show it as wrong, same as when running `nuxi typecheck`. You need to set the env `ELECTRON` as truthy for the types to match. See f492f603018691e76e5e47c4ec0a1f0778253185, where this comment was added.

Issue #162, opened 2023-06-29
Profile Settings


kkuepper:
see [Figma for how it's supposed to look](https://www.figma.com/file/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?node-id=153%3A127572&t=DM37HzaU1nHnnRjR-0).

Tasks:
- [ ] popup that can be used for dark-mode and language
- [ ] switch for dark-mode, replacing the switch introduced by #152
- [ ] switch for app-language, replacing the switch introduced by #51
- [ ] switch for content-languages
- [ ] use stored language after Refresh
- [x] adjust settings of API (see plugin `2.bmm-api`)
- [x] make sure no hard-coded uses of contentLanguage exist (e.g. Accept-Language for requests to the API)
- [x] Store auto-play setting in storage
- [x] Implement `Feedback`
- [x] Implement `Contact`
- [x] don't implement `About BMM` since it needs to be different. We'll decide exactly how at a later stage (v0.6 or later).
- [x] implement Profile dropdown (Thank you @belljohan3)

PR #169, opened 2023-07-03
Storage for content language

SimonSimCity:
Based on https://github.com/bcc-code/bmm-web/pull/165

@sclausendk We have to define what should happen if a user changes the content-language. Should we just do nothing (easy, but not intuitive), or should we reload the website (might loose state) or just try to refresh all the content on the current website (could become quite complex..)?


sclausendk:
@SimonSimCity Ideally we would not reload the page, but change the setting and provide a notification that enables the user to reload themselves?
Is a reload needed in order to reload the data?

SimonSimCity:
> Is a reload needed in order to reload the data?

@sclausendk no, or depends if you mean reloading the data or the entire website. If we do not want to do a reload of the entire website, we can either wait on the user to do an action which requires the data to reload (navigating back and forth, rerun the search, ...), or adjust our data-fetching to reload the data (not the entire website). An example to reloading just the data is the language of the website. We do not reload the entire website when changing the language, but we "reload" each text.

sorenclausenpodimo:
@SimonSimCity If thats the case, I would be completely fine with not doing anything. The user changes the language in settings, and that it.

SimonSimCity:
Solves two of the parts of #162

kkuepper (approved):

PR #170, opened 2023-07-04
Event modifier

SimonSimCity:
Make use of [event-modifiers](https://vuejs.org/guide/essentials/event-handling.html#event-modifiers) instead of writing it out in code.


SimonSimCity:
@adelinn could you have a look at why the `Deploy to Azure Static Web Apps` pipeline fails?

sifferhans (approved):

kkuepper (approved):

PR #180, opened 2023-07-21
Updated translation

SimonSimCity:
Updated translation and removed `F` button from album and playlists


locales/en.json
```diff
@@ -8,13 +8,49 @@
...
+    "logout": { "title": "Sign out" }
+  },
   "error": {
     "page-not-found": {
       "headline": "Error",
       "text": "Sorry. We can't find the page \"{path}\"."
     }
   },
+  "collection": {
+    "track-count": "{count} track | {count} tracks",
```
kkuepper, line 33: not sure if the translators will get that

locales/en.json
```diff
@@ -8,13 +8,49 @@
...
     "search": "Search"
   },
+  "sidebar": {
+    "playlists": {
+      "title": "Playlists"
+    }
+  },
+  "profile": {
+    "title": "Profile",
+    "autoplay": { "title": "Auto Play" },
```
kkuepper, line 18: I'm not sure if I like the title part. Why not call it autoplay directly?

kkuepper: Unless you give a good reason, we should change it.
It's been done differently everywhere else

kkuepper (changes requested):

locales/en.json
SimonSimCity: What format do you currently use for pluralization ..?

kkuepper: we have 2 entries: track-count-plural, track-count-single

SimonSimCity: Uff... Means you're ignoring the plural rules of languages you tend to support? https://www.unicode.org/cldr/charts/43/supplemental/language_plural_rules.html#Cardinal-Integer

You're at least ignoring it for Romanian and Polish (not mentioning the others with a much smaller user base like Russian and Slovenian)

Well, in the app the number of tracks is anyways always shown as "tracks" in English, so I see: "1 tracks" 😁

SimonSimCity: I've looked a bit more into pluralization in `vue-i18n` and it seems we will need quite some additional work to support any standardized form of language pluralization. We'll even have to define non-english pluralization rules; but they've implemented a common way of writing your own message compiler (https://github.com/intlify/vue-i18n-next/pull/1497) which might help us here. It might be that one - using this - could implement pluralization for [ICU messages](https://unicode-org.github.io/icu/userguide/format_parse/messages/) but it might be hard for e.g. [i18next](https://www.i18next.com/misc/json-format), because they use separate messages.

locales/en.json
SimonSimCity: I did it because I expect a description to be added, as we have it on the app. To mark to the translator that this is a title, having a more detailed description, I did it this way.

locales/en.json
SimonSimCity: As `vue-i18n` [published a new version](https://github.com/intlify/vue-i18n-next/releases/tag/v9.3.0) including the feature of writing a custom message compiler, we can go ahead implementing a compiler which would accept ICU messages by [following the example put out in the documentation](https://vue-i18n.intlify.dev/guide/advanced/format.html#message-compiler-implementation).

kkuepper: We decided to try the format with pipe

PR #181, opened 2023-07-21
Update plugin for typed-routing

SimonSimCity:


sifferhans:
@SimonSimCity Have you checked out the `experimental.typedRouter` option in Nuxt?

SimonSimCity:
> @SimonSimCity Have you checked out the `experimental.typedRouter` option in Nuxt?

Nope, and it doesn't seem to be documented... I guess you mean [`experimental.typedPages`](https://nuxt.com/docs/api/configuration/nuxt-config#typedpages), which I've tried earlier, but it didn't have what we needed back then or wasn't stable enough, can't remember. I'll consider switching once its stable. Feel free to set it up earlier if you feel like it.

kkuepper (approved):

PR #182, opened 2023-07-22
Feat/make carousel items overflow

adelinn:


layouts/default.vue
```diff
@@ -17,7 +17,7 @@ const { currentTrack } = inject(MediaPlaylistInjectionKey)!;
     class="flex h-full max-h-[100vh] overflow-y-hidden bg-background-1 text-black-1 dark:bg-background-dark-1 dark:text-white-1"
   >
     <SidebarElement />
-    <main class="relative flex-grow overflow-y-auto">
+    <main class="relative flex-grow overflow-y-auto" style="overflow-x: hidden">
```
sifferhans, line 20: Can we use a tailwind class here instead of inline styles?:)
See the `overflow-y-auto` class already present

pages/index.vue
```diff
@@ -81,6 +81,10 @@ watch(
         <div
           v-if="group.header?.useCoverCarousel"
           class="flex space-x-8 overflow-x-auto scrollbar-hide"
+          style="
+            padding: 0 calc((100vw - 100%) / 2) 0 calc((100vw - 100%) / 2);
```
sifferhans, line 85: What is this calculation? Could you perhaps add a screenshot to visualize it?

sifferhans (changes requested):

layouts/default.vue
SimonSimCity: I'd avoid an overflow of `x`, if possible. Often this makes things like overlays or similar effects just more complicated.

pages/index.vue
SimonSimCity: Btw ... what is actually meant by _viewport_? Is it the width of the browsers viewport, or is it the next scrollable area, which therefore yields as viewport..? Or is this even left open for browsers to interpret? 😅

On https://caniuse.com/viewport-units I found the following note:

> Currently all browsers but Firefox incorrectly consider 100vw to be the entire page width, including vertical scroll bar, which can [cause a horizontal scroll bar](https://codepen.io/CiTA/pen/zYBmYBJ) when overflow: auto is set.

... ouch 😓 ... that means you can equally use `999em` here instead of the calculation, because `100vw` will include the width of the left menu. I'd rather go for either a separate layout or [inverting the padding of the parent-element](https://stackoverflow.com/a/76748609/517914).

SimonSimCity:
@adelinn Thanks for your code contribution! There are multiple ways of solving the problem you're pointing at here. I'd like to have a chat about the various ways and implications before merging one here. Would be nice if you could make time for this to find the best possible solution.

pages/index.vue
u12206050: All this is possible with only css. Avoid using viewport as that is the entire width of the screen and only makes sense on the root elements or overlays.

Use flex to set out your initial layout eg. Sidebar and main and set shrink-0 on sidebar with width appropriately.

On the main you would then add overflow-hidden and give it a child with overflow-scroll

The main should have width-full even though it will be shrieked to accommodate the sidebar.

Depending on if the sidebar is collapsible or not you would switch between relative and absolute.

The carousel would then be width-full and overflow-x-auto with each item also being width-full with some padding unless you want to see the next item slightly in which case you might do width-4/5

SimonSimCity:
@adelinn In a discussion with Søren, he developed a new design which now does not require scrolling anymore. See https://github.com/bcc-code/bmm-web/issues/191#issuecomment-1685915058

I'll close this PR in favor of #194. Thanks a lot for your help and time also to discuss this issue!

PR #185, opened 2023-08-14
Calculate package version from git in pipeline

Ghostbird:
My Debian got confused when I tried to install a newer package, but the version number was the same as an older one. Then I realised that we don't track the version very well.

This is a proposed solution. It uses `git describe --tags` which outputs the tag followed by the commit count since that tag, and a short form of the last commit hash. It transforms such a value to a semver number using regex replace. For the current commit it yields `0.0.3.49`. The tag is `v0.0.3` and there have been 49 commits since that tag.

This of course will not generate unique versions cross-branches, or in case of history rewrites. I thought about using a timestamp instead. However, what I like about this method is that in general the final number is not going to be longer than 4 digits, and it is very easy for a human to see which of two versions is later.

Currently I've set it up that it only actually performs the action in the build pipelines just after the install. I have not added it to the `postinstall` because then it would run locally too. That's something we can discuss. In the pipeline it will write the version number in the `package.json`'s `version` value before the build. This ensures the version number is correctly set during e.g. electron packaging.

The actual version value in the `package.json` is not used any more except when developing locally.

To allow calculation of the version I've added `git fetch --unshallow` to the build pipeline. This is of course not ideal, and we may want to find a better way to do this, but so far I've not found it yet. `git fetch --tags` doesn't work.

Note that there's no sanity checking of the version number. If someone tags the repo with a non-conforming tag, the output of `git describe --tags` will be used unaltered. If you tag the repo with `0.0.4` instead of `v0.0.4`, and then add a commit before publishing, the deployment will use something like `0.0.4-1-gd04fc60` as version string, instead of `0.0.4.1`.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -31,6 +31,8 @@ jobs:
 
       - name: Install dependencies
         run: pnpm install
+      - name: Calculate package version
+        run: git fetch --unshallow && pnpm update:version
       - name: Run linters
         run: pnpm lint
       - name: Test and create coverage
--- a/.github/workflows/electron.yml
+++ b/.github/workflows/electron.yml
@@ -29,6 +29,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      - name: Calculate package version
+        run: git fetch --unshallow && pnpm update:version
+
       - name: Download certificate file
         run: |
           echo $APPLE_DEVELOPER_ID_FILE | base64 --decode >> developer-id.privatekey.p12
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "e2e": "cypress open",
     "lint": "stylelint \"**/*.{vue,scss}\" && prettier . --check && eslint --ext ts,js,vue,json . && ELECTRON=true nuxi typecheck",
     "test": "vitest",
-    "postinstall": "nuxt prepare"
+    "postinstall": "nuxt prepare",
+    "update:version": "node update-package-version.js"
   },
   "dependencies": {
     "@auth0/auth0-vue": "^2.2.0",
--- a/update-package-version.js
+++ b/update-package-version.js
@@ -1,0 +1,29 @@
+#!/usr/env node
+import { readFileSync, writeFileSync } from "fs";
+import { execSync } from "child_process";
+
+writeFileSync(
+  "package.json",
+  `${JSON.stringify(
+    {
+      ...JSON.parse(readFileSync("package.json")),
+      version: execSync("git describe --tags")
+        .toString()
+        .trim()
+        .replace(
+          /^v([0-9]+)(?:\.([0-9]+))?(?:\.([0-9]+))?(?:-([0-9]+))?(?:-g([0-9a-z]+))?.*$/,
+          (_, major, minor, patch, commitCount, objectIdentifier) =>
+            [
+              `${parseInt(major ?? "0", 10)}.${parseInt(minor ?? "0", 10)}.${
+                // Increment patch by 1 if we're not exactly at a tag.
+                // For discussion see: https://github.com/bcc-code/bmm-web/pull/185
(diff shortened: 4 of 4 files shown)


Ghostbird:
> Invalid version: "0.0.3.50"  failedTask=build stackTrace=Error: Invalid version: "0.0.3.50"

Apparently this is not a valid version number. I don't know why. Must it be only three numbers? Then maybe it will be fixed if we retag to `v0.1`.

SimonSimCity:
SemVer seems to allow only three fields splitted by dots, but allows for pre-release identifiers, separated by a dash. They just seem to always have a lower priority than the actual version 🙃 which is the opposite in our case, but this shouldn't matter all to much, IMO.

Ghostbird:
I wrote something similar once for .NET, which uses a `major.minor.patch.build` format, and that won't fly with node. I've now updated it to use `major.minor.patch-commitCount-objectIdentifier`. The latter parts are added only if `git describe --tags` generates them.

## Problem

As @SimonSimCity points out, this isn't correct semantic versioning. In our case `0.1.0-1-6800e4e` comes after `0.1.0`. However in semver `0.1.0-1-6800e4e` is the _lower_ version number. Therefore whenever there is a `commitCount` part, I'll increment the patch number by one.

However, this isn't perfect either. I don't know about other environments, but Debian doesn't implement semver's pre-release scheme and `0.0.2-1-6800e4e` comes _after_ `0.0.2`.

Therefore, I now have made a PR that nicely automatically assigns semver correct package versions on build. However in doing so I have _not_ properly solved the original issue that lead me to create this PR. At first glance it works. If I download the latest build on Debian and try to install it, it will treat is as a newer version and replace the current one. However, when the next version is released, Debian will treat that version as _older_ than the already installed pre-release, and will not install the actual newer version.

## Examples

- Build tag `v0.0.4`: version will be `0.0.4`
- Build tag `v0.1`: version will be `0.1.0`
- Tag `v0.1` and add one commit `6800e4e`, version will be `0.1.1-1-6800e4e`
- Tag `foobar`: Version will be `foobar`
- Tag `0.0.4` (missing the v) and it will seem correct: version will be `0.0.4`
- Tag `0.1` (missing the v and omitting the patch number)  and add one commit `6800e4e`, version will incorrectly be `0.1-1-6800e4e`.

##‌ Conclusion

I think that this PR may add value, but alas it doesn't provide a perfect solution. So far the only issue that I know will occur will happen to testers on Debian, but testers on other platforms _might_ be affected too. As far as I can see however, it only applies to people who test a pre-release build, and the solution is to either uninstall the pre-release before installing the release or to force install the release

SimonSimCity:
Increasing the patch-version by 1 would fit the semver style, which node uses internally for verifying, but I don't know how the different app-stores are dealing with this. This IMO would be the factor outweighing everything in this decision.

I don't know which stores we want to support after all. Just found this [in the SnapStore documentation](https://snapcraft.io/docs/revisions):
> Neither the revision number (nor its version) enforce an order of release. The local system will attempt to install whatever snap is recommended by the publisher in the channel being tracked.

[The app store seems to only support a version number with three sections](https://developer.apple.com/help/app-store-connect/update-your-app/create-a-new-version/), but they have [a different concept for beta versions](https://developer.apple.com/documentation/xcode/distributing-your-app-for-beta-testing-and-releases).

[Windows store also requires the version number to have only three sections (fourth is reserved for store-internal usage)](https://learn.microsoft.com/en-us/windows/apps/publish/publish-your-app/package-version-numbering?pivots=store-installer-pwa). This yields for PWA and MSIX. Installations via MSI or EXE do not have any limitation to the version number.

Don't know what a good strategy would be here. I thought we could even have every master-commit increase the patch-version and only tag the minor and major versions 🙃 I don't know what's a good strategy here, and would leave this up to you @kkuepper to decide. Don't know if the stores I've listed here are the ones we want to publish our app on...

Ghostbird:
I've made the assumption that we do _not_ publish pre-releases to any user-facing channels. The pre-release tags will at worst affect testers.

Issue #194, opened 2023-08-22
Implement new home-screen (without horizontal scrolling)


SimonSimCity:
A new design for the home-screen was developed, which does not require horizontal scrolling:
https://www.figma.com/file/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?type=design&node-id=1004-5327&mode=design&t=C1OFoO4SaSTXt7mh-11

If the assignee choses not to implement the first section containing "Fra Kåre" and "continue listening", he must create a separate ticket for it. Otherwise this ticket is done when the design on the home-screen is approved by Søren or Karsten.

victorienfotsing:
https://github.com/bcc-code/bmm-web/pull/196

victorienfotsing:
![image](https://github.com/bcc-code/bmm-web/assets/19670750/364c41eb-3ae9-464f-8534-134cab35cf6c)
![image](https://github.com/bcc-code/bmm-web/assets/19670750/529d207f-a185-4d42-bcd5-a68e1f2b7591)

SimonSimCity:
@victorienfotsing could you take the time looking into what's left to make the home-screen look well?

SimonSimCity:
@sclausendk Can you check if the home page is well implemented or what's yet missing? I know that the first section (Fra Kåre podcast, continue listening, ...) is yet to be implemented (I need to create a ticket for this ...), and the meta-data to tracks should look like a table (https://github.com/bcc-code/bmm-web/issues/263) - but is the rest fine?

sclausendk:
@SimonSimCity Where do I find the link to this update?

SimonSimCity:
@sclausendk it's all at the link you sent 🤗

BeastiW:
Hi @SimonSimCity  @victorienfotsing

I tried to convert it to grid/ subgrid - I thought this was an interesting task. 
Please have a look at this branch: https://github.com/bcc-code/bmm-web/tree/263-alignTrackItem-Design

I guess I need some help concerning two points:
- The mouseover on simple tracks is not working any more. When I left the empty "position: abolute" div, it only played the last track in the list... somehow the div spanned too much space or the layer of the click event is not set properly.
- to extend tailwind.conf.ts I had to extend of "figma2tailwind" template in convertToTailwind.ts (maybe this is corrent but it seems dirty).

sclausendk:
Here's a few things we need to fix:
[...]

PR #196, opened 2023-09-12
Implement new home-screen #194

victorienfotsing:
- Remove horizontal scroll
- Add see all button
- Responsive change the number of items shown
- add [@vueuse library](https://vueuse.org/)
- Update button properties


pages/index.vue
```diff
@@ -81,19 +86,34 @@ const playItem = (item: TrackModel, group: IDiscoverableGroup) => {
...
+              >
+                {{ group.header.title }}
+              </a>
+              <span v-else>{{ group.header.title }}</span>
+            </div>
+            <NuxtLink v-if="group.header.link" :to="group.header.link">
+              <ButtonStyled intent="secondary" size="small">
+                <span
+                  class="overflow-hidden text-ellipsis whitespace-nowrap text-center"
+                  >{{ t("track.a11y.see-all") }}</span
```
SimonSimCity, line 103: Please take this text separate from the `a11y` context. As this text is visible, its not an accessibility-related reason we're having this translated.

Btw... I doubt many translators do have an idea what `a11y` means, so if you'd rather be about to find a new keyword to put those translations to, go ahead.

pages/index.vue
```diff
@@ -81,19 +86,34 @@ const playItem = (item: TrackModel, group: IDiscoverableGroup) => {
...
+                :href="group.header.link"
+              >
+                {{ group.header.title }}
+              </a>
+              <span v-else>{{ group.header.title }}</span>
+            </div>
+            <NuxtLink v-if="group.header.link" :to="group.header.link">
+              <ButtonStyled intent="secondary" size="small">
+                <span
+                  class="overflow-hidden text-ellipsis whitespace-nowrap text-center"
```
SimonSimCity, line 102: Could we remove the `overflow-hidden` here? Is it actually needed after all when having `whitespace-nowrap` and `text-ellipsis` in place, doesn't this contradict each other?

Actually, what about removing all of the mentioned? Shouldn't this button be to be read in its full text?

SimonSimCity (changes requested):
Since you've removed the last real use of `scrollbar-hide`, please remove also the extension which adds this option to the codebase. It's also mentioned in the component `components/playlist/PlaylistCarousel.vue` but this component is not in use and can be removed also.

pages/index.vue
victorienfotsing: In which context should the translation have to be added?

pages/index.vue
victorienfotsing: whitespace-nowrap is required to keep the button height for small screen. But the other classes can be removed

pages/index.vue
SimonSimCity: I would add it to a `home.list.see-all` or something... The button is IMO not really related to tracks... otherwise I have another ticket where I'm taking up some things about translation, where I could pick this up later 👍

SimonSimCity:
Resolving the outstanding problem the linter is pointing at is more crucial to me than the things mentioned in the comments here. I could also tackle those things later..

victorienfotsing:
see #202
this PR will be closed

PR #197, opened 2023-09-13
search box view

belljohan3:
Building the search view so that it matchs with the design


SimonSimCity (approved):
Looks good 😎
Changes which could be added:
* Translation
* Creating a component for the tabs
* Technical implementation of the search

Just for future additions: I'd like a visual addition, which easily can be removed if a feature is not yet ready. One example would be that a button will turn red on hover; another is to add a comment in the code.

PR #202, opened 2023-09-16
Feature/implement new home screen

victorienfotsing:


victorienfotsing:
#194 #196

package.json
```diff
@@ -16,6 +16,7 @@
     "coverage": "vitest run --coverage",
     "e2e": "cypress open",
     "lint": "stylelint \"**/*.{vue,scss}\" && prettier . --check && eslint --ext ts,js,vue,json . && ELECTRON=true nuxi typecheck",
+    "lint:fix": "stylelint --fix \"**/*.{vue,scss}\" && prettier . --check --write && eslint --fix --ext ts,js,vue,json .",
```
SimonSimCity, line 19: Could you please check first what combining the options `--check` and `--write` does to the prettier command?

SimonSimCity (commented):
Looks good 😎 Please confirm about the lint-fix command you've added.

package.json
victorienfotsing: I have removed the option `--check` since it is useless for lint fix

Issue #153, opened 2023-06-26
Refresh access token before it expires


kkuepper:
After the app has been open for more than 24h, I can't play anymore. I'm pretty sure that happens because the access token is valid for 24h and has expired.
As a work-around you can close and reopen the app, which gets you a new access token.

SimonSimCity:
Something which might be helpful here: https://stackoverflow.com/a/66604849/517914

SimonSimCity:
Changed the title because it's not only electron where this needs to be fixed

SimonSimCity:
Should be fixed by #204. If its not, we can reopen this ticket and add some details to it.

PR #204, opened 2023-09-16
Refresh access token

Ghostbird:
To solve #153

I haven't managed to test this properly yet. I tried to connect this to my own Auth0 client, to set a shorter access token time-out. That didn't work well, because that access token doesn't grant access to BMM. So you get access denied anyway.

So now I hope this PR‌ will make a build, and I can install that locally. Then I'll use that that build for a few days, to see whether it works.


SimonSimCity:
Thanks for the separate commits, now I could just cherry-pick it and leave out the unrelated changes 🤗

PR #208, opened 2023-09-16
Search composable file created

belljohan3:


SimonSimCity:
Would be nice if you could add this to the place where this function is used. I'd wait for this PR before merging this.

belljohan3:
@SimonSimCity what do you mean when you say the place ?

SimonSimCity:
@belljohan3 currently this PR would add some code which is not used. The changes look good, but they do not add a value yet, because the code you added is not used. I'd merge this PR when you add this function and call it somewhere where the user can use it 😊

SimonSimCity:
Am merging this now. It's already adding some value which we haven't had yet. Additional improvements on this can be provided in other PRs and you're very welcome and encouraged to provide additional value.

belljohan3:
Ok @SimonSimCity. Thanks a lot.

PR #209, opened 2023-09-16
only call setAuthenticatedUserContext once we have the personId

kkuepper:


SimonSimCity (approved):
LGTM 👍 By this we avoid sending a user-context of "unknown", but rather skip adding a user-context to our analytics tool if the age of the person visiting is unknown

PR #210, opened 2023-09-16
Fix excessive type assertions

Ghostbird:
Just a minor refactor. Only assert the type once, when the object is created, instead of on every use.


Ghostbird:
This one was cherry picked and will probably fail until rebased on #211

SimonSimCity:
Since this was made in the early beginning, and I don't yet know which compile process brought me to do this, I'd rather leave it as is instead of wasting much time to test the different states the app compiler can be in. It worked before, and I don't really see a reason to pick it up again. I might check this later though, but not now.

PR #211, opened 2023-09-16
Disable `no-param-reassign` for properties.

Ghostbird:
The primary purpose of `no-param-reassign` is to guard the behaviour of `arguments`[^1]. Apparently one of our defaults has additionally enabled the sub-rule for argument property reassignment. I don't think it's a good fit for a Vue + Nuxt project as those APIs make it necessary in some methods. E.g. `defineNuxtPlugin` may have to modify properties of the input `nuxtApp`.

[^1]: https://eslint.org/docs/latest/rules/no-param-reassign

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,6 +25,7 @@
         "prettier"
       ],
       "rules": {
+        "no-param-reassign": ["error", { "props": false }],
         "@typescript-eslint/consistent-type-assertions": [
           "warn",
           { "assertionStyle": "never" }


SimonSimCity:
For a change in type rules, I'd like to spend a bit more time on going through what this means for the project. For this reason, I'll not merge it yet.

SimonSimCity:
I'd refrain from merging this. Reassigning values to parameters of an argument is not considered a good practice in general, because it can - if not controlled - quickly lead to bugs which are very hard to track down. Since this project will be open to people which are on quite different skill levels, I say that a rule like this should be generally disapproving such code, while still giving more experienced developers the option to show that they know what they're doing. But that's my opinion. @kkuepper what's your objection here?

kkuepper:
I guess we're all on the same page that 
```
var foo = function(bar) {
    bar = 13;
}
```
should not be allowed.
Do I understand it correctly that it's mainly about "props": true vs "props": false?

Personally, I think "props": false is better. As @Ghostbird points out there will definitely be cases where it's necessary to do that and then there shouldn't be a rule annoying you about it.

Ghostbird:
Yes, there's no question about the normal `no-param-reassign`. That rule makes perfect sense. As the PR title indicates, this is about applying it to _properties_ of arguments, which is what `props: true` does.

I'd be happy to leave `props: true`, because it could catch unintended errors, were it not for the problematic fact that we use APIs that conflict with it, as demonstrated above. Therefore I'd rather disable it, than keep it enabled, since it introduces legitimate doubt about the applicability of linter rules. I'd rather work with a situation were linter rules are always correct.

SimonSimCity:
Thanks for the honest discussion. I see your reason and agree with it. I'll merge, thanks 😎🎉👍

PR #220, opened 2023-09-22
Bug/183 cannot scroll when dropdown is opened

patricktamal:
I just use the native dropdown that comes with Tailwind instead custom DropDown Component that on my point is not appropriate for this usage.


SimonSimCity:
Just for clarification: This does not come with tailwind but with headlessui 🤗

PR #230, opened 2023-11-02
update nuxt and core packages

sifferhans:
- update Nuxt and some of its related packages
	- `nuxt`: 3.6.1 -> 3.8.0
	- `@nuxt/kit`: 3.6.1 -> 3.8.0
	- `@nuxt/devtools`: 0.6.4 -> 1.0.0
- fix a couple of imports

Mainly to stay kinda up-to-date with Nuxt releases, but Nuxt 3.8 also has some nice additions such as client-side caching for `useAsyncData` calls (if that could be relevant)
https://nuxt.com/blog/v3-8


nuxt.config.ts
```diff
@@ -72,4 +72,11 @@ export default defineNuxtConfig({
       },
     ],
   },
+  typescript: {
```
SimonSimCity, line 75: Since I couldn't find any justification - why was this option added? The documentation states that this is not recommended (https://nuxt.com/blog/v3-8#type-import-changes), and also the typescript documentation (https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) looks like it was a good addition to have this on.

sifferhans: I originally didn't have this enabled, but the amount of changed files grew quite a bit when going through and fixing the imports. I disabled it to keep the PR as small as possible, but I can enable it again and fix the imports if that is better?

SimonSimCity: The current amount of PRs this would get in conflict with is comparably small - almost none, if I see it correctly. For the matter of keeping PRs small, those changes are all related to one reason, which is updating while keeping good code settings, which takes higher priority for me here. Having this setting enabled improves how well tree-shaking will work, which should reduce the amount of code we load in production. So if not yet, when? 😎

sifferhans:
Yeah, it seems like I keep unintentionally messing with the lock file 😂

sifferhans:
@SimonSimCity Looks like eslint caught some imports it doesn't like. I can see if I can fix it

SimonSimCity:
We might have to postpone these changes, because I see that it creates some conflicts with electron - the app doesn't start, at least not in DEV mode.

sifferhans:
@SimonSimCity I see 👍🏼

SimonSimCity:
With the latest updates, it works most of the time. In development, the web-app just crashed on the first start with an error message. Most of the sources I found was because Vue seems to be run twice, but I couldn't wrap my head around how this could be and whether this could also happen in production...
Screenshot 2023-12-08 at 12 57 46

This could also be something with Electron having to start up twice since the last set of updates...

sifferhans:
@SimonSimCity Does Electron have to start twice?

SimonSimCity:
@sifferhans I have no idea which mechanism this is to blame on - just have noticed this after updating... It's just for dev though, not for a built electron application.

kkuepper:
Feel free to reopen this PR after the Frivillighetshelg.

SimonSimCity:
Finally a version which doesn't slow down anymore, but updates everything to its latest compatible version 🎉😃

kkuepper (approved):
looks good to me. Can we merge it?

PR #232, opened 2023-11-03
Update all packages except the nuxt packages

SimonSimCity:
In this PR, I've updated all packages except those updated in PR #230 and our i18n package (see PR #205). Everything seems to look good as far as I've been able to test.


SimonSimCity:
The main difference for the coverage report is, that the new stats include all files, whereas the old only includes those files which are imported by `.spec.ts` files, hence the worsening in "lines of code" and the other stats.

SimonSimCity:
@sifferhans It seems that there was a problem with the prettier configuration for quite some time now, which is why it didn't run in the latest version of vscode on my machine either. I've now fixed the prettier-config in the repo and now it runs just fine. Can you please check if this PR fixes the issue with prettier for you as well?

sifferhans:
@SimonSimCity Will take a look!

SimonSimCity:
@sifferhans merged it now since I had this in quite many of the other branches also. This should resolve the issue of prettier not running.

PR #233, opened 2023-11-04
Show error of login module

SimonSimCity:
Errors by the login module (like about a expiry token missing, introduced by #204), were not reported to the user but silently hidden.

This change will show the login error to the user, will give him the option to log out and try again (which is what helps for #204) or contact the support (mail address taken from the profile-menu).


Ghostbird:
Thanks for fixing my mess!

Ghostbird (approved):

PR #236, opened 2023-11-04
Feature: Create private playlist

SimonSimCity:
@sclausendk there was no design of how this should look like, so I created one my own which to me doesn't look all to bad 😎 Would be nice to get your feedback on this. The icons are just stolen from somewhere I found fitting ...


sclausendk:
I cant really see what you have done, since I can only follow along on the alpha site 👀

sifferhans (approved):

kkuepper (approved):

PR #243, opened 2023-11-18
Add pinia-plugin-persistedstate dependency and

lukaspoloki:
configure persisted state options


SimonSimCity (changes requested):
At least the changes in the file `tsconfig.json` and the store for the toolbar title should be looked at. The settings are nothing I'd bother much about, but I just prefer sticking to the defaults if necessary.

.vscode/settings.json
SimonSimCity, line 1: Please remove the changes here. If you prefer to use these, that's up to you, but this shouldn't be part of the general repository.

SimonSimCity (changes requested):

SimonSimCity (approved):

SimonSimCity:
Thanks very much for this contribution 🎉 😃

Issue #237, opened 2023-11-17
Setup Nuxt plugin for Plausible (Analytics)


kkuepper:
https://github.com/nuxt-modules/plausible

<script defer data-domain="bmm-web.brunstad.org" src="https://plausible.io/js/script.js"></script>

BeastiW:
I added the Nuxt plugin in the configuration and the package.json - now we could start adding custom events like e.g. `useTrackEvent("shuffle");`

PR #249, opened 2023-11-18
237 setup nuxt plugin for plausible analytics

BeastiW:
closes [issue 237](https://github.com/bcc-code/bmm-web/issues/237)


kkuepper (approved):

kkuepper (approved):

nuxt.config.ts
```diff
@@ -31,6 +32,9 @@ export default defineNuxtConfig({
       systemName: process.env.ELECTRON ? "Electron" : "Web",
     },
   },
+  plausible: {
```
SimonSimCity, line 35: @BeastiW - what's the reason for configuring the domain here? On the documentation (https://github.com/nuxt-modules/plausible#basic-usage) I read:
> By default, @nuxtjs/plausible will use window.location.hostname for the Plausible domain configuration key, which will suit most users.

What was the reason for specifying the domain?

BeastiW: The reason: I did not think about the reason ;-) I guess window.location.hostname should be enough. I just took it from the issue description. I can remove it and create another PR

PR #253, opened 2023-11-19
Refactor to css variables

u12206050:
Updated pipeline to convert figma variables to css variables and merging light and dark mode variables.
Added styles for scrollbar


SimonSimCity:
@u12206050 can you please explain what you did here and how to do the adjustments you've done here in future development?

u12206050:
Sure, basically instead of having to set the dark setting for each Color used every time we just globally swap the css variables.

So now you can go from 
'text-tint dark:text-tint-dark' to just 'text-tint'

The same goes for 'bg'

As a bonus, not that we need it, but since they are css variables we could easily change themes live on the browser. Eg. Try changing the css variable of '--bmm-tint' to another color in your browser.

PR #254, opened 2023-11-19
Profile dialogs

kkuepper:
- Theme dialog is done
- interface language: still needs to style the select (doesn't look too bad though)
- content language: added an empty dialog


components/ProfileMenu.vue
```diff
@@ -5,16 +5,24 @@ import { Menu, MenuButton, MenuItem, MenuItems, Switch } from "@headlessui/vue";
 const profileStore = useProfileStore();
 const { t } = useI18n();
 
-const colorMode = useColorMode();
-const colorTheme = computed(() => {
-  switch (colorMode.preference) {
+const showInterfaceLanguageDialog = ref(false);
+const showContentLanguageDialog = ref(false);
+const showThemeDialog = ref(true);
```
SimonSimCity, line 10: Should this really be set to `true` ..? Won't this show the dialog when rendering this component?

SimonSimCity (changes requested):
Prettier still warns about some incompatible formatting in some files. After addressing that and the comment here, I'm fine with merging it.

components/ProfileMenu.vue
kkuepper: Ooops. I used it for easier debugging.

SimonSimCity (approved):

PR #256, opened 2023-11-23
[Dialog] Use HeadlessUI Dialog component as base of `DialogBase` component

sifferhans:
Using the HeadlessUI Dialog as a base gives us better accessibility out of the box, with keyboard navigation, close on escape, better screen reader support etc.

### DIFF
--- a/components/ChangeLocale.vue
+++ b/components/ChangeLocale.vue
@@ -1,5 +1,3 @@
-<script lang="ts" setup></script>
-
 <template>
   <select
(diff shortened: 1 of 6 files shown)


SimonSimCity:
Is there a reason not to merge it? To me, it looks good to me already now 😎

sifferhans:
@SimonSimCity I haven't implemented the Content Language dialog yet, but that can probably be a separate PR 😁

SimonSimCity (approved):

PR #257, opened 2023-11-25
Add Browse page

### LINKED ISSUE #97: Implement browse-view

sifferhans:
Closes #97

### DIFF
--- a/pages/browse/index.vue
+++ b/pages/browse/index.vue
@@ -1,13 +1,56 @@
 <script lang="ts" setup>
+import { BrowseApi } from "@bcc-code/bmm-sdk-fetch";
+
 const { t } = useI18n();
 toolbarTitleStore().setReactiveToolbarTitle(() => t("nav.browse"));
 
-const { data: playlists } = useCuratedPlaylists();
+const { data: browseSections, pending } = await useBrowse();
 </script>
 
 <template>
   <div>
     <PageHeading class="mb-6">{{ $t("nav.browse") }}</PageHeading>
-    <PlaylistOverview v-if="playlists" :playlists="playlists" />
+    <template v-if="pending">
+      <ul>
+        <li
+          v-for="index in 5"
+          :key="index"
+          class="my-6 h-11 w-full animate-pulse rounded-lg bg-background-2"
+        ></li>
+      </ul>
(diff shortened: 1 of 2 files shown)


pages/browse/index.vue
```diff
@@ -2,12 +2,53 @@
...
+        <AlbumItem v-else-if="section.type === 'album'" :album="section" />
+        <PlaylistItem
+          v-else-if="section.type === 'playlist'"
+          :playlist="section"
+        />
+        <PodcastItem
+          v-else-if="section.type === 'podcast'"
+          :podcast="section"
+        />
+      </template>
```
SimonSimCity, line 51: Please add a v-else which marks the other possible cases as red (not implemented) or something, or make somehow sure that we'll be informed by type hinting or similar that there is no other possible case. If you want to leave this up to me, please add a `Todo` in the code.

SimonSimCity (changes requested):
Looks good - just one addition to make this more future proof, so we'll be informed when things are about to break 😎

pages/browse/index.vue
```diff
@@ -2,12 +2,54 @@
...
+      </ul>
+    </template>
+    <ul v-else>
+      <template v-for="section in browseSections" :key="section.id || 0">
+        <PageHeading v-if="section.type === 'section_header'" :level="3">
+          <div class="flex items-center justify-between">
+            <div>
+              <NuxtLink
+                v-if="typeof section.link === 'string'"
+                :to="section.link as any"
```
SimonSimCity, line 27: Do we need this `as any` here? The nuxt-link seems to anyways have some trouble with external links in the desktop-app, since they're external links.

And - wouldn't this be a duplicate of the next `a`-tag?

SimonSimCity:
Two additional things I've overlooked last time 😅

pages/browse/index.vue
sifferhans: This is only done because I wanted to use `NuxtLink` here incase we would use relative link paths, but if we won't I'll just change this back to an `a` tag.

sifferhans:
@SimonSimCity I've done the changes you mentioned

SimonSimCity (approved):

PR #258, opened 2023-11-27
Implement Podcast view

### LINKED ISSUE #132: View for podcasts
On the app, there's almost no difference in viewing podcasts, albums or playlists (curated or private). This could be a first step for the web-version, too. The major addition would be, that the list of tracks can be extended when reaching the bottom of the list of currently loaded items.

This page is needed for the home-screen (#64).

@sclausendk - would be nice to have a short response from you on the design.

sifferhans:


pages/playlist/podcast/[id]/index.vue
```diff
@@ -1,3 +1,77 @@
...
+  id: collectionId,
+});
+
+const trackGroups = computed(() => {
+  if (tracks.value) {
+    const sorted = tracks.value.sort((a, b) =>
+      a.publishedAt! > b.publishedAt! ? -1 : 1,
+    );
+
+    // split the tracks into groups of 7
```
Comment (SimonSimCity) [inline pages/playlist/podcast/[id]/index.vue:19]: This just says the same as the code below. Useful would instead be a reasoning why you are doing this.

If you split in groups of 7, saying those are the episodes per week, you're assuming, that there's a new episode every day of a week (not only on working days) - additionally, you're assuming that the previous week ended 7 items before, which doesn't match if someone looks this up on a Wednesday.

I'd like to discuss this a bit before going ahead.

pages/playlist/podcast/[id]/index.vue
```diff
@@ -1,3 +1,77 @@
+<script lang="ts" setup>
+const { t } = useI18n();
+toolbarTitleStore().setReactiveToolbarTitle(() => t("nav.podcast"));
+
+const { id } = useRoute<"playlist-podcast-id">().params;
+const collectionId = Number(id);
+
+const { data: podcast } = usePodcast({ id: collectionId });
+const { data: tracks, pending: tracksPending } = usePodcastTracks({
```
Comment (SimonSimCity) [inline pages/playlist/podcast/[id]/index.vue:9]: Since podcasts are basically limitless, and the API applies some sort of limit here, we should at least indicate, that this might not be the full list, but a "load more" functionality is not yet implemented.

SimonSimCity (changes requested):
Just some comments on this draft... Please let me know if I shouldn't point out all these things on a draft but rather on normal PRs 😊

pages/playlist/podcast/[id]/index.vue
```diff
@@ -1,3 +1,77 @@
...
+  id: collectionId,
+});
+
+const trackGroups = computed(() => {
+  if (tracks.value) {
+    const sorted = tracks.value.sort((a, b) =>
```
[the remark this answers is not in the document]
Comment (sifferhans) [reply pages/playlist/podcast/[id]/index.vue:15]: Yeah, thats true 😆 No need to do more work than necessary. I'll remove this

pages/playlist/podcast/[id]/index.vue
Comment (SimonSimCity) [reply pages/playlist/podcast/[id]/index.vue]: Since we're talking about grouping the podcasts by the week they were published, I'd go for something like https://stackoverflow.com/a/6117889/517914

SimonSimCity:
@sifferhans If you remove the usage of `useHead()` I'll merge it right away. The usage of `useHead()` should be introduced as one bulk to every place where we need it - it's a conceptual change rather than a change I would do just here and there.

I'd leave the related ticket as open for the next person to implement grouping by week - or maybe you want to pick it up now by grouping by calendar week (https://stackoverflow.com/a/6117889/517914)

sifferhans:
@SimonSimCity Removed `useHead`. We can merge this in now, and keep the issue open for the grouping functionality to be implemented later

SimonSimCity (approved):

SimonSimCity:
@sifferhans sorry for the remarks on `useHead`... I just see now that this has already been used at quite some other places already... Awaiting the feedback from Søren now on #260...

PR #259, opened 2023-11-27
Add contribution guide

sifferhans:
I put @kkuepper contribution issue into a Markdown file, as issues is the wrong place for that.
Should probably be rewritten with examples etc.


SimonSimCity:
I agree with putting it into the code, I do not agree with the content in there though. I'll see how I can refine this and will do this in the ticket soon. What's written in there is not a viable way of driving development as of now, IMO.

Issue #239, opened 2023-11-17
Popup for Track details / More information


kkuepper:
- [x] Implement component for modal/popup
- [x] fill popup with information about the track ([see Figma](https://www.figma.com/file/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?type=design&node-id=801-26347&mode=design&t=uI8D4WpY5cQpD42J-0))
- [x] make links work (when clicking on 3 dots)

#240 also needs the popup component. Therefore it should be implemented as a separate PR that gets merged quickly.

![Screenshot 2023-11-17 at 2 34 33 PM](https://github.com/bcc-code/bmm-web/assets/649978/368da0b9-6fc2-421f-bd69-af969b3c87e0)

SimonSimCity:
Open questions:
- [ ] Which types of references do we have?
- [ ] What locale should we use to show the published-at date? Is the locale of the current environment good enough, or should we use something related to the chosen language?

About references ... Maybe it's good to update the API to have a better typing here. There seems to be some kind of enum-type per relation of a track (https://github.com/bcc-code/bmm-api/blob/develop/BMM.Api.Core/DatabaseContract/RelationType.cs#L10) but its not typed, excepted for some typings I found (https://github.com/bcc-code/bmm-api/blob/develop/BMM.Api.Core/DatabaseContract/Track.cs#L216-L278). @kkuepper Would be good to know if its any good to improve the api here or whether at all its possible to prove that's even possible.

kkuepper:
About locale:
The current website uses "5. Jan 2024 - 14:26" which is the european version. I think it's fine to keep doing that. Do whatever is easy to do.

About references:
I'm not sure what you want from me. It sounds like the typing doesn't work properly so we need to fix it. But you (@SimonSimCity) usually do that.

SimonSimCity:
Fixed by https://github.com/bcc-code/bmm-api/pull/480, https://github.com/bcc-code/bmm-api/pull/485 and https://github.com/bcc-code/bmm-web/pull/301

PR #264, opened 2023-12-28
added missing track detail information

BeastiW:
should solve second subtask of https://github.com/bcc-code/bmm-web/issues/239 but there is an open question about "Reference" and "hör sangen her" link.

### DIFF
--- a/components/track/TrackDetails.vue
+++ b/components/track/TrackDetails.vue
@@ -1,20 +1,71 @@
 <script setup lang="ts">
 import { TrackModel } from "@bcc-code/bmm-sdk-fetch";
 
+const { t } = useI18n();
+
 defineProps<{
   track: TrackModel;
 }>();
 </script>
 
 <template>
-  <div class="py-4 text-lg">
-    {{ track.title }}
+  <div class="relative flex items-center justify-normal gap-3">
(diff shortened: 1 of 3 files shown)


components/track/TrackDetails.vue
```diff
@@ -1,20 +1,71 @@
...
+    </div>
+    <div>{{ track.meta?.album }}</div>
+  </div>
+  <hr class="bg-background-2 border-0 h-[2px]" />
+  <div class="py-4 text-lg flex">
+    <div class="w-40">
+      <b>{{ t("track.details.publish-date") }}</b>
+    </div>
+    <div>
+      {{ track.publishedAt?.toLocaleDateString() }}
```
SimonSimCity, line 41: Thanks for using a localized date representation 🎉 It's just debatable whether a user wants to see it this way when choosing another language - but that's again language vs locale - which is worth a separate discussion.

locales/de.json
SimonSimCity, line 1: Just for next time: It's not worth updating the locale-files for other languages excepted for English. All other languages are overwritten by our translation platform Crowdin.

SimonSimCity (approved):

SimonSimCity:
@BeastiW Thanks for your contribution 🎉

PR #266, opened 2024-01-08
Fix electron mac notarization

SimonSimCity:
What I did:

1. Converted configuration to ~TypeScript~ JavaScript (couldn't leave it in TypeScript because there still is [a bug in the config-compiler](https://www.github.com/mighdoll/config-file-ts/pull/22), so I left it at a jsDoc annotation), because I like having things typed and someone complain if a property is misspelled or has a wrong type.
2. Take "team-id" which was already provided. [If we would go back to YAML, we would have to define the team-id in the configuration file.](https://www.github.com/electron-userland/electron-builder/issues/7812)
3. Tried a bit back and forth until I could confirm it working.


kkuepper (approved):
The only functional change is

```
notarize: {
   teamId: process.env.APPLE_TEAM_ID || "",
},
```

Looks simple enough. Thanks for figuring it out.

PR #269, opened 2024-01-17
Update many packages

SimonSimCity:
Update quite some npm packages. As far as I've tested, everything still works as expected.

- Linting
- Electron-DEV
- Tests


components/track/TrackItem.vue
```diff
@@ -1,6 +1,6 @@
 <script lang="ts" setup>
-import type { NuxtIconName } from ".nuxt/nuxt-icons";
-import type { RoutesNamedLocations } from ".nuxt/typed-router";
+import type { NuxtIconName } from "~/.nuxt/nuxt-icons";
```
sifferhans, line 2: We should probably import from `#build/*` instead of `~/.nuxt/*` here. `#build` is an alias for the `.nuxt` directory, if I'm not mistaken

package.json
```diff
@@ -20,57 +20,58 @@
...
     "log-symbols": "^6.0.0",
-    "nuxt": "^3.6.1",
-    "nuxt-electron": "^0.6.0",
-    "nuxt-typed-router": "^3.3.2",
-    "nuxt-vitest": "^0.11.2",
-    "prettier": "^3.0.3",
-    "prettier-plugin-tailwindcss": "0.4.1",
-    "stylelint": "^15.11.0",
-    "stylelint-config-recommended-scss": "^13.1.0",
+    "nuxt": "^3.6.5",
```
sifferhans, line 62: Nuxt 3.9 is already released. Is it too big of a jump from 3.6.1?

sifferhans (approved):
Just a couple of questions, other than that looks good:)

package.json
SimonSimCity: Can't go to `3.7.0` or further as of now. See https://github.com/nuxt/nuxt/discussions/25257 and the there linked research.

components/track/TrackItem.vue
SimonSimCity: Fixed.

package.json
SimonSimCity: The update will be followed up in #230

kkuepper (commented):
looks good to me

PR #271, opened 2024-01-23
add back button

kkuepper:


components/AppToolbar.vue
```diff
@@ -1,12 +1,28 @@
+<script setup type="ts">
+const route = useRoute();
+const showBackButton = () =>  route.name !== "index" && route.name !== "search" && route.name !== "browse";
```
sifferhans, line 3: We can turn this into a `computed` for it to be updated reactively:)

sifferhans (approved):
Just a small thing. Looks good otherwise

components/AppToolbar.vue
kkuepper: done

components/AppToolbar.vue
```diff
@@ -1,12 +1,28 @@
+<script setup type="ts">
+const route = useRoute();
+const showBackButton = computed(() => route.name !== "index" && route.name !== "search" && route.name !== "browse");
```
SimonSimCity, line 3: Do we want to limit it based on the route the user is at, or at the ability of going back..?

kkuepper: In the design, the back button is missing if you're on one of the top level pages. (pages that have a dedicated button in the menu.)
For a deeplink we'd have 2 options:
- don't show the back button since there's nothing to go back to.
- show back button and navigate to home

PR #274, opened 2024-01-24
263 align track item design

### LINKED ISSUE #263: Align `TrackItem` design to Figma
The track items in the app are not correctly aligned to each other; it should be more table-like, with consistent column widths and alignment throughout the rows.

I would think this could be a nice use case for CSS subgrid. Or just using `<table>` 🤪

https://caniuse.com/css-subgrid

## Figma
CleanShot 2023-12-07 at 15 08 49

## App
CleanShot 2023-12-07 at 15 10 14

SimonSimCity:
Fix #263.

@BeastiW Extending `tailwind.conf.ts` looks good to me as of now. We do not yet have a common approach for doing it otherwise. It's hacky, but fills the need for now.

The mouseover and other things should be fixed. Would like to have your feedback on the changes I've done. Please compare the PR also against the current design in terms of spacing etc - I haven't taken too much care about this. would be nice if you could do it.

### DIFF
--- a/components/album/AlbumItem.vue
+++ b/components/album/AlbumItem.vue
@@ -7,8 +7,11 @@ defineProps<{
 </script>
 
 <template>
-  <li class="group col-span-5 relative mr-3 cursor-pointer py-3">
+  <li
+    class="group grid col-span-full relative grid-cols-subgrid cursor-pointer gap-3 py-3"
+  >
     <NuxtLink
+      class="grid col-span-full grid-cols-subgrid"
       :to="{
         name: 'album-id',
         params: { id: album.id },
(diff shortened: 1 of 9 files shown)


components/album/AlbumItem.vue
```diff
@@ -7,7 +7,7 @@ defineProps<{
 </script>
 
 <template>
-  <li class="group relative mr-3 cursor-pointer py-3">
+  <li class="group col-span-5 relative mr-3 cursor-pointer py-3">
```
sifferhans, line 10: Could we do `col-span-full` here? So we don't have to change this manually if we decide to add another column later

kkuepper:
@BeastiW looks quite good.

But I think responsiveness can be improved.

Screenshot 2024-01-24 at 11 07 50 AM
The title should get more space than singsong/speech in this case


Screenshot 2024-01-24 at 11 05 56 AM
If you have a lot of space, the content should be limited and centered and not fill the whole screen.
[See Figma for details](https://www.figma.com/file/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?type=design&node-id=153-127570&mode=design&t=lDwKhvxoXnvAjIQR-0)


But I think it's fine to merge this PR since it's clearly an improvement and we could just create an issue with the things I outlined since they are a new problem.

SimonSimCity:
@kkuepper I applied now a width of `auto` to all columns, excepted the first (because we know how wide a cover will be). This should fix your concern. If not we'll cover it in another PR. I'll merge this as is for now.

PR #275, opened 2024-01-24
extract shared code into DocumentList

kkuepper:


kkuepper:
@SimonSimCity I'm still working on implementing the "See all" pages. But you can already review this part.

pages/browse/events.vue
```diff
@@ -0,0 +1,29 @@
+<script lang="ts" setup>
+import type { IAllDocumentModels } from "@bcc-code/bmm-sdk-fetch";
+
+const { data: models, pending } = useBrowseEvents();
+
+toolbarTitleStore().setReactiveToolbarTitle(() => models.value?.title ?? " ");
```
kkuepper, line 6: @SimonSimCity @sifferhans 
The title is translated on the server.
But that means I don't have a title until I get the response from the server. How can I set the toolbar title?

components/DocumentList.vue
```diff
@@ -0,0 +1,40 @@
...
+  IAllDocumentModels,
+} from "@bcc-code/bmm-sdk-fetch";
+
+const props = defineProps<{
+  type?: "page" | "playlist" | "private-playlist";
+  list: DocumentListIAllDocumentModels | null;
+  pending: boolean;
+}>();
+
+toolbarTitleStore().setReactiveToolbarTitle(() => props.list?.title ?? " ");
```
kkuepper, line 14: @SimonSimCity @sifferhans
The title is translated on the server.
But that means I don't have a title until I get the response from the server. How can I set the toolbar title?
(Moved it from another file. So you might have gotten this comment twice)

pages/browse/events.vue
kkuepper: moved it to DocumentList.vue

components/DocumentModels.vue
```diff
@@ -0,0 +1,150 @@
+<script setup lang="ts">
+import type { IAllDocumentModels, TrackModel } from "@bcc-code/bmm-sdk-fetch";
+import type { IDiscoverableGroup } from "~/composables/discover";
+import { breakpointsTailwind, useBreakpoints } from "@vueuse/core";
+
+const { t } = useI18n();
+
+const props = withDefaults(
+  defineProps<{
+    type?: "page" | "playlist" | "private-playlist";
```
SimonSimCity, line 10: This property doesn't seem to be used anywhere. Is it needed?

composables/discover.ts
```diff
@@ -10,7 +10,7 @@ export type IDiscoverableGroup = {
   items: Exclude<IAllDocumentModels, SectionHeaderModel>[];
 };
 
-export function useDiscover(requestParameters: DiscoverGetRequest) {
+export function useDiscoverOld(requestParameters: DiscoverGetRequest) {
```
SimonSimCity, line 13: This does not seem to be used anywhere... Can it be removed?

components/DocumentModels.vue
```diff
@@ -0,0 +1,150 @@
...
+
+const playItem = (item: TrackModel, group: IDiscoverableGroup) => {
+  const items = group.items.filter((c): c is TrackModel => c.type === "track");
+  setQueue(
+    items,
+    items.findIndex((track) => track.id === item.id),
+  );
+};
+
+// this is only needed while it's hosted on bmm-web.brunstad.org (Actually, it might be needed for )
```
SimonSimCity, line 51: If you mean Electron, then this might be counter-productive... See https://github.com/bcc-code/bmm-web/blob/main/electron/main.ts#L38-L44

components/DocumentList.vue
SimonSimCity: Needs to be added on the title-store (https://github.com/bcc-code/bmm-web/blob/main/stores/toolbarTitle.ts) but then we should also do some more extended handling of the watcher to properly stop it when another title is set. Otherwise the title might be weirdly overwritten by a previous setting of the title when the language is changed.

pages/index.vue
```diff
@@ -71,83 +25,7 @@ const isSmallScreen = breakpoints.smallerOrEqual("lg");
...
-            ></ContributorListItem>
-            <li v-else>
-              <div style="background-color: rgba(255, 0, 0, 0.4); color: red">
-                "{{ item.type }}" is not yet implemented ...
-              </div>
-            </li>
-          </template>
-        </ol>
-      </template>
+      <DocumentModels v-if="models !== null" :models="models"></DocumentModels>
```
SimonSimCity, line 28: When including the few lines above, this seems to be replacible by `<DocumentList>`. Or was there a reason not to do it, which I overlooked?

pages/index.vue
```diff
@@ -1,67 +1,21 @@
...
+const parameters: DiscoverGetRequest = {
+  // Compatibility is ensured in i18n.config.ts file
+  lang: locale.value as LanguageEnum,
 };
-
-const breakpoints = useBreakpoints(breakpointsTailwind);
-
-const isSmallScreen = breakpoints.smallerOrEqual("lg");
+if (userData.age) parameters.age = userData.age;
+const { data: models, pending } = useDiscover(parameters);
```
SimonSimCity, line 13: Having this not wrapped in a `watch` will not load this again when the language is changed. In my opinion, we should keep the behavior, otherwise it looks like the headlines of the home screen do not adjust the language when changed, but the rest does.

SimonSimCity (changes requested):
I've now reviewed it as if it was a normal PR 😅 Apart from these changes, I'd merge it straight away. Codewise I only see an improvement, and some conceptual arrangements which someone not familiar with deeper concepts would have to read himself into (not for time-to-time contributors).

components/DocumentModels.vue
kkuepper: removed

composables/discover.ts
kkuepper: removed

components/DocumentModels.vue
kkuepper: changed the behavior

pages/index.vue
kkuepper: true.

I actually also need to add that it's translated on the server. Those endpoints still use an older concept where the server only sends the translation id.

pages/index.vue
kkuepper: <DocumentModels ...> is supposed to be used in a lot of cases. E.g. all pages under browse use it to render the documents.
So the page needs to handle loading of documents and DocumentModels renders them.

kkuepper:
@SimonSimCity I also need to update bmm-sdk-fetch since I found 2 bugs in the definition when working with this PR. Until that happens the linter will fail.

components/DocumentModels.vue
```diff
@@ -0,0 +1,149 @@
...
+  const items = group.items.filter((c): c is TrackModel => c.type === "track");
+  setQueue(
+    items,
+    items.findIndex((track) => track.id === item.id),
+  );
+};
+
+// We remove the hostname so that we use SPA links (without full page refresh)
+const parseLink = (link: string) => {
+  const url = new URL(link);
```
SimonSimCity, line 47: Are all URLs absolute by definition? If the URL is not absolute, this line will throw an error `Uncaught TypeError: Failed to construct 'URL': Invalid URL`.

SimonSimCity:
@kkuepper Currently the linter fails because:

> components/DocumentModels.vue(121,17): error TS2367: This comparison appears to be unintentional because the types '"track" | "album" | "playlist" | "podcast" | "InfoMessage" | "Tile" | "contributor" | "listening_streak" | "recommendation" | "tile_message" | "tile_video" | "track_collection" | "year_in_review"' and '"chapter_header"' have no overlap.

The reason for `chapter_header` not being valid option on the `type` property is because it already has been filtered out - it can by definition not be part of this list anymore (see `composables/discover.ts`).

If you want, we can elaborate about it - as I don't quite know the concept you're trying to set here, I guess - that's at least what I read off the responses you wrote to my comments ...

components/DocumentModels.vue
kkuepper: yes. They are absolute.

kkuepper:
> @kkuepper Currently the linter fails because:
> 
> > components/DocumentModels.vue(121,17): error TS2367: This comparison appears to be unintentional because the types '"track" | "album" | "playlist" | "podcast" | "InfoMessage" | "Tile" | "contributor" | "listening_streak" | "recommendation" | "tile_message" | "tile_video" | "track_collection" | "year_in_review"' and '"chapter_header"' have no overlap.
> 
> The reason for `chapter_header` not being valid option on the `type` property is because it already has been filtered out - it can by definition not be part of this list anymore (see `composables/discover.ts`).
> 
> If you want, we can elaborate about it - as I don't quite know the concept you're trying to set here, I guess - that's at least what I read off the responses you wrote to my comments ...

Updating bmm-sdk fixed that issue.

components/DocumentList.vue
kkuepper: done

pages/index.vue
kkuepper: I added UiLanguage as Http header so that the Api knows the language for every request.
So sending the language as QueryParameter is no longer needed (e.g. see Discover request).

kkuepper:
@SimonSimCity @sifferhans It's ready for final review / merge.

sifferhans (approved):
Looks good:)

components/DocumentList.vue
```diff
@@ -0,0 +1,44 @@
...
+
+const props = defineProps<{
+  list: DocumentListIAllDocumentModels | null;
+  pending: boolean;
+}>();
+
+watch(
+  props,
+  () => {
+    toolbarTitleStore().setToolbarTitle(props.list?.title ?? " ");
```
SimonSimCity, line 15: Do all pages who use this for listing set this as title?

kkuepper: Not sure what you mean. If the server returns a DocumentList it also provides the title.

components/DocumentList.vue
```diff
@@ -0,0 +1,44 @@
...
+    </template>
+    <template v-else>
+      <DocumentModels
+        v-if="
+          list !== null &&
+          list !== undefined &&
+          list.items !== null &&
+          list.items !== undefined
+        "
+        :models="list.items as IAllDocumentModels[]"
```
SimonSimCity, line 40: Type assertion should not be necessary here. Why was it used?

_Note:_ Type assertion should be avoided whenever possible. See https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions - you'll loose all benefit of a type-safe system. This is valid Typescript:
```ts
const foo = 123;
const bar: string = foo as any as string;
```

composables/browse.ts
```diff
@@ -5,3 +5,27 @@ export function useBrowse() {
     useLazyAsyncData("browse", () => new BrowseApi().browseGet()),
   );
 }
+
+export function useBrowseEvents() {
+  return reactiveApi(
+    useLazyAsyncData("browse", () => new BrowseApi().browseEventsGet()),
```
SimonSimCity, line 11: If the result of this API call differs from `BrowseApi().browseGet()`, it should have another key - not `browse`. Otherwise the caching algorithm will most likely create some problems in the future 🙈

components/DocumentModels.vue
```diff
@@ -0,0 +1,149 @@
+<script setup lang="ts">
+import type { IAllDocumentModels, TrackModel } from "@bcc-code/bmm-sdk-fetch";
+import type { IDiscoverableGroup } from "~/composables/discover";
```
SimonSimCity, line 3: This type should be moved in here since it's not used in the composable anymore.

composables/playlist.ts
```diff
@@ -35,3 +35,11 @@ export function useCuratedPlaylists() {
     useLazyAsyncData("playlists", () => new PlaylistApi().playlistGet()),
   );
 }
+
+export function useFeaturedPlaylists() {
+  return reactiveApi(
+    useLazyAsyncData("playlists-featured", () =>
```
SimonSimCity, line 41: Just hope nobody names his playlist `featured` or we'll get in conflict with the caching of the function in line 11... maybe best to use a different caching-id there...

pages/browse/audiobooks.vue
```diff
@@ -0,0 +1,12 @@
+<script lang="ts" setup>
+import type { DocumentListIAllDocumentModels } from "@bcc-code/bmm-sdk-fetch";
+
+const { data: models, pending } = useBrowseAudiobooks();
+</script>
+
+<template>
+  <DocumentList
+    :list="models as DocumentListIAllDocumentModels"
```
SimonSimCity, line 9: Is the type assertion necessary here?

pages/browse/events.vue
```diff
@@ -0,0 +1,12 @@
+<script lang="ts" setup>
+import type { DocumentListIAllDocumentModels } from "@bcc-code/bmm-sdk-fetch";
+
+const { data: models, pending } = useBrowseEvents();
+</script>
+
+<template>
+  <DocumentList
+    :list="models as DocumentListIAllDocumentModels"
```
SimonSimCity, line 9: Is the type assertion necessary here?

pages/browse/index.vue
```diff
@@ -17,36 +16,8 @@ const { data: browseSections, pending } = useBrowse();
...
-        />
-        <PodcastItem
-          v-else-if="section.type === 'podcast'"
-          :podcast="section"
-        />
-        <p v-else style="color: red">This is not implemented yet</p>
-      </template>
-    </ul>
+    <template v-else>
+      <DocumentModels v-if="models !== null" :models="models"></DocumentModels>
```
SimonSimCity, line 20: Why not use the component `DocumentList` here?

components/DocumentModels.vue
kkuepper: done

composables/browse.ts
kkuepper: copy paste error. done

pages/browse/music.vue
```diff
@@ -1,3 +1,12 @@
+<script lang="ts" setup>
+import type { DocumentListIAllDocumentModels } from "@bcc-code/bmm-sdk-fetch";
+
+const { data: models, pending } = useBrowseMusic();
+</script>
+
 <template>
-  <h1>The route {{ useRoute().name }} has not been implemented yet.</h1>
+  <DocumentList
+    :list="models as DocumentListIAllDocumentModels"
```
SimonSimCity, line 9: Is the type assertion necessary here?

pages/browse/podcasts.vue
```diff
@@ -1,3 +1,12 @@
+<script lang="ts" setup>
+import type { DocumentListIAllDocumentModels } from "@bcc-code/bmm-sdk-fetch";
+
+const { data: models, pending } = useBrowsePodcast();
+</script>
+
 <template>
-  <h1>The route {{ useRoute().name }} has not been implemented yet.</h1>
+  <DocumentList
+    :list="models as DocumentListIAllDocumentModels"
```
SimonSimCity, line 9: Is the type assertion necessary here?

pages/featured.vue
```diff
@@ -1,3 +1,12 @@
+<script lang="ts" setup>
+import type { DocumentListIAllDocumentModels } from "@bcc-code/bmm-sdk-fetch";
+
+const { data: models, pending } = useFeaturedPlaylists();
+</script>
+
 <template>
-  <h1>The route {{ useRoute().name }} has not been implemented yet.</h1>
+  <DocumentList
+    :list="models as DocumentListIAllDocumentModels"
```
SimonSimCity, line 9: Is the type assertion necessary here?

composables/playlist.ts
kkuepper: it uses id, which is an integer. Also one uses singular whereas the other plural "playlist".

pages/index.vue
```diff
@@ -71,86 +22,7 @@ const isSmallScreen = breakpoints.smallerOrEqual("lg");
...
-            ></ContributorListItem>
-            <li v-else>
-              <div style="background-color: rgba(255, 0, 0, 0.4); color: red">
-                "{{ item.type }}" is not yet implemented ...
-              </div>
-            </li>
-          </template>
-        </ol>
-      </template>
+      <DocumentModels v-if="models !== null" :models="models"></DocumentModels>
```
SimonSimCity, line 25: Why not use the component `DocumentList` here?

stores/toolbarTitle.ts
```diff
@@ -11,5 +11,9 @@ export const toolbarTitleStore = defineStore("toolbarTitle", () => {
     );
   }
 
-  return { setReactiveToolbarTitle, toolbarTitle };
+  function setToolbarTitle(title: string) {
+    toolbarTitle.value = title;
+  }
+
+  return { setReactiveToolbarTitle, toolbarTitle, setToolbarTitle };
```
SimonSimCity, line 18: This implementation is prone to the problem, that the title will be overwritten if the reactive title is updated late.

pages/browse/audiobooks.vue
kkuepper: It's not necessary. Nice!

SimonSimCity (changes requested):

pages/browse/events.vue
kkuepper: fixed

pages/browse/index.vue
kkuepper: I don't understand the question

pages/browse/music.vue
kkuepper: fixed

pages/browse/podcasts.vue
kkuepper: fixed

components/DocumentList.vue
SimonSimCity: Well... I'd rather put these lines into each page... For me, the title is a part of the page and should not be controlled by a component. By this, you can use this component at more pages.

components/DocumentList.vue
kkuepper: fixed.

pages/featured.vue
kkuepper: fixed

utils/reactiveApi.ts
```diff
@@ -1,8 +1,11 @@
-import type { AsyncData } from "#app";
+import type { AsyncData } from "nuxt/app";
 
 export default function reactiveApi<Data, Error>(data: AsyncData<Data, Error>) {
   const stopHandler = watch(
-    () => contentLanguageStore().contentLanguages,
+    [
+      () => useNuxtApp().$i18n.locale,
```
kkuepper, line 6: that probably doesn't work

SimonSimCity (approved):

PR #279, opened 2024-02-05
Fix layout

kkuepper:
I adjusted the layout to better reflect what's in Figma.
Originally I just wanted to give the expanded player some fixed space. But since that caused some side-effects in other areas, I needed to fix more and more.


SimonSimCity (approved):
LGTM from the code - can't judge every effect in detail..

PR #281, opened 2024-02-08
Add track to a playlist

kkuepper:


components/track/TrackMenu.vue
```diff
@@ -7,16 +7,23 @@ import { Menu, MenuButton, MenuItem, MenuItems } from "@headlessui/vue";
 const { t } = useI18n();
 const { addNext, addToQueue } = useNuxtApp().$mediaPlayer;
 
-defineProps<{
-  track: TrackModel;
-}>();
+withDefaults(
+  defineProps<{
+    track: TrackModel;
+    buttonClass: string;
```
SimonSimCity, line 13: Have I overlooked something, or is this parameter currently unused?

components/track/TrackAddToPlaylist.vue
```diff
@@ -0,0 +1,58 @@
...
+const selectedPlaylistId = ref(0);
+const idsToLink = (...ids: number[]) => ids.map((id) => `</track/${id}>,`);
+
+const selectList = (playlistId: number) => {
+  selectedPlaylistId.value = playlistId;
+  const options: TrackCollectionIdPostRequest = {
+    id: playlistId,
+    link: idsToLink(props.trackId),
+  };
+  new TrackCollectionApi().trackCollectionIdPost(options);
```
SimonSimCity, line 24: I'd rather hide this logic in a composable for the playlist. This way, we could also move the function `idsToLink` there, so this component can fully focus on the things related to what's visible and not about the API connection.

components/track/TrackMenu.vue
SimonSimCity: Overlooked the use because I was not searching for a dash-case variant of the name 😅

components/track/TrackAddToPlaylist.vue
SimonSimCity: Fixed in https://github.com/bcc-code/bmm-web/pull/293, also added better error handling.

PR #282, opened 2024-02-08
persist and restore window position

kkuepper:
this uses ElectronStore to persist the window bounds in a json file and restore them at next app open.

@steffanhalv Would be nice to profit from your experience if that's enough or if other platforms need additional code. It seems to work fine on my M1 Mac when adding or removing monitors.


steffanhalv:
I also only have a mac with m1 chip now, but I think that should do the job.

Issue #289, opened 2024-02-09
go to contributors


kkuepper:
- [x] read contributors from track.contributors
- [x] Add button for each contributor. e.g. "Go to Kåre J. Smith", "Go to Dag Helge Bernhardsen"
- [x] navigate to "/playlist/contributor/{contributorId}"

philipdalen:
@kkuepper do we have a design for this?

kkuepper:
The design is already implemented. It just doesn't do anything.

<img width="321" alt="Screenshot 2024-02-09 at 6 55 34 PM" src="https://github.com/bcc-code/bmm-web/assets/649978/c43b37df-6612-4429-ab71-b06e58f14fa6">

PR #295, opened 2024-02-10
Add navigation for going to contributors profile

philipdalen:
#289

### DIFF
--- a/components/track/TrackContributors.vue
+++ b/components/track/TrackContributors.vue
@@ -13,7 +13,10 @@ defineProps<{
     <div v-for="(contributor, index) in track?.contributors" :key="index">
       <NuxtLink
         class="rounded-lg text-label-1 flex flow-row gap-3 p-2 px-5 hover:bg-background-3"
-        :to="`/playlist/contributor/${contributor.id}`"
+        :to="{
+          name: 'playlist-contributor-id',
+          params: { id: contributor.id ?? 0 },
+        }"
         @click="showMediaPlayer = false"
       >
         {{ contributor.name }}
--- a/components/track/TrackContributors.vue
+++ b/components/track/TrackContributors.vue
@@ -1,0 +1,23 @@
+<script setup lang="ts">
+import type { TrackModel } from "@bcc-code/bmm-sdk-fetch";
+
+const showMediaPlayer = inject<boolean>("showMediaPlayer");
+
+defineProps<{
+  track: TrackModel;
+}>();
+</script>
+
+<template>
+  <div class="bg-background-2 rounded-lg">
+    <div v-for="(contributor, index) in track?.contributors" :key="index">
+      <NuxtLink
+        class="rounded-lg text-label-1 flex flow-row gap-3 p-2 px-5 hover:bg-background-3"
+        :to="`/playlist/contributor/${contributor.id}`"
+        @click="showMediaPlayer = false"
+      >
+        {{ contributor.name }}
+      </NuxtLink>
+    </div>
+  </div>
+</template>
--- a/components/track/TrackContributors.vue
+++ b/components/track/TrackContributors.vue
@@ -1,8 +1,6 @@
 <script setup lang="ts">
 import type { TrackModel } from "@bcc-code/bmm-sdk-fetch";
 
-const showMediaPlayer = inject<boolean>("showMediaPlayer");
-
 defineProps<{
   track: TrackModel;
 }>();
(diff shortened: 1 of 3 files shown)


components/track/TrackMenu.vue
```diff
@@ -62,7 +64,20 @@ const dropdownMenuItemsForTrack = (track: TrackModel) => {
...
+    clickFunction: () => {
+      if (track.contributors && track.contributors?.length > 1) {
+        showContributorsList.value = true;
+      } else if (
+        track.contributors &&
+        track.contributors.length === 1 &&
+        track.contributors[0]?.id
+      ) {
+        navigateTo(`/playlist/contributor/${track.contributors[0].id}`);
+        if (showMediaPlayer) {
```
kkuepper, line 76: why does clicking on a contributor affect the player?

components/track/TrackMenu.vue
```diff
@@ -62,7 +64,20 @@ const dropdownMenuItemsForTrack = (track: TrackModel) => {
   items.push({
     icon: "icon.person",
     text: t("track.dropdown.go-to-contributors"),
-    link: { name: "browse" }, // TODO: change link
+    clickFunction: () => {
+      if (track.contributors && track.contributors?.length > 1) {
```
SimonSimCity, line 68: Optional chaining shouldn't be needed when you've checked its existence. Since you're using optional chaining, you just need one check. The second condition could also be simplified by using the optional chaining on the array index.

components/track/TrackContributors.vue
```diff
@@ -0,0 +1,23 @@
...
+  track: TrackModel;
+}>();
+</script>
+
+<template>
+  <div class="bg-background-2 rounded-lg">
+    <div v-for="(contributor, index) in track?.contributors" :key="index">
+      <NuxtLink
+        class="rounded-lg text-label-1 flex flow-row gap-3 p-2 px-5 hover:bg-background-3"
+        :to="`/playlist/contributor/${contributor.id}`"
```
SimonSimCity, line 16: Please use an object-structure for navigation instead. This will avoid dead-links as we by type-safety can ensure those links actually exist in the router.

components/track/TrackMenu.vue
```diff
@@ -62,7 +64,20 @@ const dropdownMenuItemsForTrack = (track: TrackModel) => {
...
+        showContributorsList.value = true;
+      } else if (
+        track.contributors &&
+        track.contributors.length === 1 &&
+        track.contributors[0]?.id
+      ) {
+        navigateTo(`/playlist/contributor/${track.contributors[0].id}`);
```
SimonSimCity, line 75: Please use an object-structure for navigation instead. This will avoid dead-links as we by type-safety can ensure those links actually exist in the router.

SimonSimCity (changes requested):
Please fix the urls.

components/track/TrackMenu.vue
philipdalen: Because the intention when clicking on the contributor is to view the profile of the contributor, I assume, having the player open then covers up the screen so you have to minimize it manually. I just do it automatically which I thought is more expected behaviour but I might be wrong?

components/track/TrackMenu.vue
philipdalen: done

components/track/TrackContributors.vue
```diff
@@ -0,0 +1,23 @@
...
+defineProps<{
+  track: TrackModel;
+}>();
+</script>
+
+<template>
+  <div class="bg-background-2 rounded-lg">
+    <div v-for="(contributor, index) in track?.contributors" :key="index">
+      <NuxtLink
+        class="rounded-lg text-label-1 flex flow-row gap-3 p-2 px-5 hover:bg-background-3"
```
kkuepper, line 15: hover:bg-background-3 looks weird since it gets white. I would expect a darker color. I don't think Søren has provided a color yet, though.

components/track/TrackContributors.vue
philipdalen: done

components/track/TrackContributors.vue
kkuepper: Søren just added that it should be hover:bg-label-separator

components/track/TrackMenu.vue
kkuepper: The idea is that the player claims the right side of view and is not on top of anything. I believe on smaller screen sizes it does cover some content since the content has a minimum width.

Anyway, I don't think we should hide the player

components/track/TrackContributors.vue
philipdalen: done

philipdalen: done

components/track/TrackMenu.vue
philipdalen: Done

SimonSimCity (approved):

Issue #65, opened 2023-04-21
design for player


kkuepper:
- [x] make expanded player fill right side of the screen (available space for content is reduced)
- [x] Animate playing icon in queue (blocked due to missing design from Søren)
- [x] Implement transition (see prototype)
- [x] Marquee for the meta-data
- [x] Let user drag the progress-bar on mouse-down, update the time while keeping the song playing, and update the players position only on mouse-up. (Moved to #305)

a lot has already been implemented. Those 5 tasks are what's left.

- demo: https://www.figma.com/proto/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?node-id=488-16188&scaling=min-zoom&page-id=488%3A16147
- figma: https://www.figma.com/file/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?node-id=488-16188&t=uTTeaIsiJ23rNNzX-0

![Image](https://user-images.githubusercontent.com/649978/233626051-0e6731af-181f-47b9-9079-467b256a00f4.png)

kkuepper:
The first draft of the mini player has been implemented by @ZDerekh but the bigger version still has to be done.

Joaveinz:
What is meant by the first draft @kkuepper @SimonSimCity ?

Is this what I can currently see in the Main branch?

Joaveinz:
Simon just explained to be that what I see now in the main branch is the First Draft 👍

Joaveinz:
@SimonSimCity @kkuepper What is the difference between these two designs of the media player? Has a final design be decided?

<img width="1138" alt="image" src="https://github.com/bcc-code/bmm-web/assets/54300414/8e94536d-5b67-41b0-92e2-b48e8f69c84b">

Joaveinz:
Simon has informed me that we go with the one on the right, as that's the one that he's seen being used more often.

Joaveinz:
@SimonSimCity and @kkuepper  Here's a small update since last week


Closed player:
![image](https://github.com/bcc-code/bmm-web/assets/54300414/7501f900-5821-46f8-8d67-94d94caa2fdf)


On hover:
![image](https://github.com/bcc-code/bmm-web/assets/54300414/982626c4-7c39-41a6-bc74-d610f801e6f3)


Open Player:
![image](https://github.com/bcc-code/bmm-web/assets/54300414/68970b64-6345-4479-a8eb-3afb9805e084)


[...]

PR #296, opened 2024-02-10
65 design for player

BeastiW:
#65  fixes all remaining checkmarks except of the last bullet point.


SimonSimCity (approved):

SimonSimCity:
The marquee of the text is quite good, but there are still cases where it could do the wrong thing, e.g. when resizing the page (maybe also when changing orientation on a tablet). As of now, I'd say that's rather over-engineering - but if anyone has a better approach, I'd be happy to take it.

Does this PR also solve the marquee of meta-data in the queue and in the mini-player?

kkuepper:
if the text doesn't need scrolling, it's now left aligned. Is there a way to make it centered?

kkuepper (changes requested):
looks quite distracting if it's not centered

kkuepper (approved):
looks awesome! Well done!

PR #299, opened 2024-02-10
Feature/fix design comments

kkuepper:


components/ProtectedImage.vue
```diff
@@ -1,10 +1,16 @@
...
 
-const props = defineProps<{
-  src: string;
-  alt?: string;
-}>();
+const props = withDefaults(
+  defineProps<{
+    src: string;
+    alt?: string;
+    noBorder: boolean;
```
SimonSimCity, line 8: If it is optional, it should be marked with a question mark here; otherwise the system thinks its required.

components/ProtectedImage.vue
```diff
@@ -19,5 +25,15 @@ watch(
 );
 </script>
 <template>
-  <img :src="source" :alt="alt || ''" loading="lazy" v-bind="$attrs" />
+  <img
+    :src="source"
+    :alt="alt || ''"
+    loading="lazy"
+    v-bind="$attrs"
```
SimonSimCity, line 32: I know it was used before... but if this is for the `class` keyword, this can be removed.

SimonSimCity (commented):
Have to go - will check the rest later.

components/ProtectedImage.vue
kkuepper: done

components/ProtectedImage.vue
kkuepper: done

PR #304, opened 2024-02-11
Better marquee

SimonSimCity:
Added a component for content in marquee, but also changed the marquee-effect to be linear, but first start after 2 seconds. Furthermore, the effect will start over again if a new text appears, which also is too long.

The space between the repeated text is `pr-14`, which is added in the `marquee` component if the variable `contentIsTooLarge` is true.

The mini-player does not have the text as marquee (as discussed @kkuepper), but the full-size player has it both for the title and the second line.

The speed is not a fixed amount of seconds, but the `marquee` component has a speed defined in pixel per second.

I also found it quite annoying if a text, which just didn't quite wrap by very few pixel (not noticeable) also was rotating, which is why I added `bufferSpace`. I just have an example for when the title in the mini-player still rotated - could be removed if requested.

This component also readjusts if the effect is not needed anymore due to resizing (fixing https://github.com/bcc-code/bmm-web/pull/296#issuecomment-1937003918).

### DIFF
--- a/components/MediaPlayer.vue
+++ b/components/MediaPlayer.vue
@@ -70,14 +70,14 @@ const onPointerDownProgressBar = () => {
           <div
             class="flex gap-1 w-full flex-col overflow-hidden whitespace-nowrap"
           >
-            <TextMarquee>
+            <div class="w-full truncate">
               <h3
                 class="truncate text-lg font-semibold leading-tight"
                 :title="currentTrack?.title || ''"
               >
                 {{ currentTrack?.title }}
               </h3>
-            </TextMarquee>
+            </div>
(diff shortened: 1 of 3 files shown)


BeastiW (approved):
I think it is nice - in my opinion it could maybe be a bit slower (maybe 30 px/sec?)

kkuepper:
@BeastiW approved the changes and what's deployed to https://kind-meadow-01b17e903-304.westeurope.3.azurestaticapps.net/ looks very good.
As far as I'm concerned @SimonSimCity needs to decide if he wants to make it slower as @BeastiW or merges it as is.

SimonSimCity:
30px/s looks better, updated.

PR #306, opened 2024-02-14
finetune search

kkuepper:


SimonSimCity:
I'd rather revert the changes you've done in 7734ed613621a033c3a2be9a0d407b20ee15542d. The reason I'd do it is because a user cannot differentiate between a page which is not yet implemented (it should have something when we release, even if it's a text saying "Not yet implemented" just not in red). I'm otherwise quick to assume, that there's some kind of swallowed exception in the code if the page is just empty.

kkuepper:
It's not an empty page. It redirects you to the right url. That way urls from the current website still work while also making sure people use the new url in the future. Try it: https://kind-meadow-01b17e903-306.westeurope.3.azurestaticapps.net/playlist/curated/91/Daily%20Mix

pages/playlist/private/[id]/[...name].vue
```diff
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-const { id } = useRoute<"playlist-private-id">().params;
+const { id } = useRoute<"playlist-contributor-id-name">().params;
```
Comment (kkuepper) [inline pages/playlist/private/[id]/[...name].vue:2]: contributor seems wrong

Comment (SimonSimCity) [reply pages/playlist/private/[id]/[...name].vue]: Thanks! Fixed.

SimonSimCity:
LGTM, can be merged. I'll just later want to enable the eslint-rule `@intlify/vue-i18n/no-dynamic-keys` to allow us to check programmatically if all translations are enabled...

SimonSimCity:
I've tried something regarding to what you've worked on. In one way, it looks like we can use https://nuxt.com/docs/guide/directory-structure/pages#keepalive to make this work - there are just some issues here and there (e.g. https://github.com/vuejs/core/issues/8509, which has a work-around (https://www.github.com/vuejs/core/issues/5603#issuecomment-1715369413)).

I'd have to research this properly, but I wouldn't spend too much time on it as of now 😎 If someone wants to go for it, this can be done via looking at this PR again when taking it up again.

PR #307, opened 2024-02-15
Better playing animation

SimonSimCity:
Updated the morphing of the playing animation and linked the playing-state to the paused-state of the player.


kkuepper (approved):
Looks great.
For the future though: Soren provided a lottie file with a playing animation: https://github.com/bcc-code/bmm-app/blob/ad0c79877c9b0609621c6b30458ec11ca06c21f3/BMM.UI.Android/Resources/raw/play_animation_icon.json

If we do further changes it should be about using that.
But for now I would merge this and focus on other things.

PR #317, opened 2024-02-22
implement route /track/{id}

kkuepper:


SimonSimCity:
I've added a loading indicator and some form of error handling.

SimonSimCity:
The player does not start, and I assume it's some kind of initialization that doesn't happen properly. Wasn't able to find out right now what it was - would have to dig deeper into it... Looks good otherwise 😅

kkuepper:
> The player does not start, and I assume it's some kind of initialization that doesn't happen properly. Wasn't able to find out right now what it was - would have to dig deeper into it... Looks good otherwise 😅

I think the browser blocks the player from starting without any user interaction (to prevent websites being annoying). The old website has the same problem. I think Google added that in 2019 or so.

PR #319, opened 2024-02-22
implement repeat

kkuepper:


plugins/mediaPlayer/mediaPlayer.ts
```diff
@@ -114,6 +127,8 @@ export const initMediaPlayer = (
 
         if (hasNext.value) {
           next();
+        } else if (queue.value.isRepeatEnabled) {
+          restart();
```
SimonSimCity, line 131: Is it correct to show that there's no next track in the list? "HasNext" will be falsy when having reached the end of the list.

kkuepper: good point. changed it

kkuepper:
@SimonSimCity I also added that previous() jumps to the beginning if we're more than 5s into the track.

SimonSimCity:
@kkuepper before merging, I want to have some time to make a thorough walk-through of these changes to ensure they do not create any other downsides. The logic of the player is quite complex and sometimes even depends on things which are not fully covered by the documentation which - without tests, can easily create a lot of situations where this does not work as intended.

I'm thinking of reworking the repeat-option fully into the playlist, so the media-player only needs a minimal amount of changes. I think this would make the addition of this feature more stable than it is now.

As for the issue mentioned by you, I'd rather put it into a separate PR if possible.

SimonSimCity:
You've added a feature, but only changed tests? 😅 That sounds suspicious to me ... That means that features are not working as expected? I'll check and create a PR at least for reverting the adjustments to tests and create new ones.

plugins/mediaPlayer/mediaPlayer.ts
```diff
@@ -85,6 +90,19 @@ export const initMediaPlayer = (
...
+  function play() {
+    if (activeMedia.value) {
+      activeMedia.value.play();
+    } else {
+      initCurrentTrack();
+    }
+  }
+
+  function restart() {
+    if (queue.value.length === 1) play();
```
SimonSimCity, line 102: I doubt that this actually has worked, because it doesn't reset the ended-flag of the track. I've refactored this in a new PR where I'm also implementing the single-repeat.

PR #320, opened 2024-02-22
add and delete playlists

kkuepper:


components/PrivatePlaylistMenu.vue
```diff
@@ -0,0 +1,184 @@
...
+    id: props.playlist.id,
+  });
+  refreshPrivatePlaylists();
+};
+
+const dropdownMenuItems = () => {
+  const items: DropdownMenuItem[] = [];
+
+  items.push({
+    // icon: "icon.ai", // ToDo: get comment icon
```
kkuepper, line 73: @SimonSimCity @sifferhans I think Søren added that icon fairly recently. How do I get new icons?

SimonSimCity: You only have to add a file to the `assets/icons` folder. The nuxt-plugin should pick up that you added a file and update the typescript definition.

kkuepper: done

PR #322, opened 2024-02-23
don't send age in requests anymore

kkuepper:


pages/index.vue
```diff
@@ -2,11 +2,9 @@
 import type { DiscoverGetRequest } from "@bcc-code/bmm-sdk-fetch";
 
 const { t } = useI18n();
-const userData = useNuxtApp().$userData;
 toolbarTitleStore().setReactiveToolbarTitle(() => t("nav.home"));
 
 const parameters: DiscoverGetRequest = {};
```
SimonSimCity, line 7: I'd remove this line as well then and pass in an empty object if needed. Looks good otherwise, even though I don't know the reason why this was done ...

kkuepper: Previously we didn't have the age (birthdate) available on the server. Therefore the client had to send it. We've changed that now and then it makes a lot of sense to not send it as part of the url.

PR #323, opened 2024-02-24
:recycle: code review fixes for PR 313

BeastiW:


stores/contentLanguage.ts
SimonSimCity, line 1: We should revert this, because it can be a source of various problems in the future. Since the content of this store is permanently saved, we cannot ensure its value will always stay compatible to the enum, which might change with the next release of the API.

SimonSimCity:
I've added some additional feedback on #313.

SimonSimCity (approved):
LGTM 😎 I'll take a closer look at this functionality as a whole unit and maybe create a PR if other things jump into my eye.

kkuepper (approved):

PR #327, opened 2024-02-26
Feature/error logging

kkuepper:


app.vue
```diff
@@ -1,21 +1,48 @@
...
       await loginWithRedirect();
     }
   },
   { immediate: true },
 );
 
+watch(error, async () => {
+  const e = error.value;
+  if (e) {
+    const errorCode = (e as any).error;
```
SimonSimCity, line 25: I wouldn't cast this to `any` but rather check if it has the properties we expect or whether it is an instance of the class we expect.

app.vue
```diff
@@ -1,21 +1,48 @@
...
     if (!isAuthenticated.value) {
+      $appInsights.event("auth0 - redirect to login", {});
       await loginWithRedirect();
     }
   },
   { immediate: true },
 );
 
+watch(error, async () => {
+  const e = error.value;
```
SimonSimCity, line 23: Can be received as first argument of the function you give as parameter to `watch` - see how we do for `isLoading`.

app.vue
```diff
@@ -1,21 +1,48 @@
...
+  if (e) {
+    const errorCode = (e as any).error;
+    console.error(e.message, "\n\n\n", errorCode);
+    $appInsights.event("auth0 - error", {
+      error: errorCode,
+      message: e.message,
+      stack: e.stack,
+    });
+    if (errorCode === "missing_refresh_token") {
+      // For these errors we've identified that a logout & relogin likely will fix it.
```
SimonSimCity, line 33: Don't you want this to be logged differently in this case, since we have a way of solving it?

utils/reactiveApi.ts
```diff
@@ -8,5 +8,18 @@ export default function reactiveApi<Data, Error>(data: AsyncData<Data, Error>) {
     },
   );
 
+  watch(data.error, async () => {
```
SimonSimCity, line 11: I'd do this differently; specially because this has nothing to do with being reactive, and therefore, some of the API calls are not wrapped by this function.

I'd either create a separate wrapper or see whether we can integrate this with some way of error-handling the API-package already has in store.

SimonSimCity (changes requested):

SimonSimCity:
Related to #33

app.vue
kkuepper: I'm not sure if it solves the issue. I first need some real world data. It solved it when I manually provoked the issue on my machine. But there's a good chance it doesn't work in other circumstances.

To get good logging requires an iterative approach. I'll probably create 6-10 small PRs with very small changes over the next 3 months to get this to a good state.

app.vue
kkuepper: done

app.vue
kkuepper: done

utils/reactiveApi.ts
kkuepper: once the bmm-sdk correctly calls onError we can remove it.

SimonSimCity: My point here also is a different one. By this, you're not catching the errors of all API calls, just those wrapped into this function. Some are not wrapped into this function.

kkuepper: I'm aware of that. That's why it's only temporary until onError works. But 80% of errors is better than 0%.

PR #341, opened 2024-03-08
fix: media player auth token was not being refreshed

andreasgangso:
**Problem**
Right now if you leave the tab open until the access token expires, it's not going to refresh the token and playback won't start.
**Cause**
The mediaplayer had an unecessary cache for the token. It's not being refreshed upon expiry because we are watching "isAuthenticated".
**Solution**
The auth0 sdk has an internal cache, so we can use getAccessTokenSilently() directly.
This also mirrors how it's done everywhere else, e.g. https://github.com/bcc-code/bmm-web/blob/0cb1e1d9915236d1e0d66c943015abc253aa2081/components/ProtectedImage.vue#L16

### DIFF
--- a/plugins/mediaPlayer/mediaPlayer.ts
+++ b/plugins/mediaPlayer/mediaPlayer.ts
@@ -35,12 +35,11 @@ export interface MediaPlayer {
   replaceCurrent: (track: TrackModel) => void;
 }
 
-export const authToken = ref<string | undefined>();
-
 export const seekOffset = 15;
 
 export const initMediaPlayer = (
   createMedia: (src: string) => HTMLAudioElement,
+  getAccessToken: () => Promise<string>,
   appInsights: AppInsights,
   user: IUserData,
 ): MediaPlayer => {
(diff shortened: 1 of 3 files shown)


plugins/mediaPlayer/mediaPlayer.ts
```diff
@@ -84,9 +84,7 @@ export const initMediaPlayer = (
       nextStartPosition = 0;
     }
     const token = await getAccessToken();
-    activeMedia.value = new MediaTrack(
-      createMedia(authorizedUrl(url, token)),
-    );
+    activeMedia.value = new MediaTrack(createMedia(authorizedUrl(url, token)));
```
kkuepper, line 87: It's the same code. The syntax is slightly different, but I can't find any change

andreasgangso: are you looking at only the latest commit?

kkuepper: oops

kkuepper:
Looks good but it breaks the existing tests

kkuepper (changes requested):
tests need to be fixed

andreasgangso:
The linked PR is about cloudfront tokens which is something else.

kkuepper (approved):

SimonSimCity:
Nice, thanks for the addition 🎉 I've just brought it one level further in #342 😊

PR #342, opened 2024-03-09
Don't pospone track initialization

SimonSimCity:
In PR #341 a mechanism was introduced which refetches the access-token before the start of every track (if not cached - see [`getAccessTokenSilently()`](https://auth0.github.io/auth0-vue/interfaces/GetTokenSilentlyOptions.html)).

Since `getAccessTokenSilently()` returns a promise, this would delay the instantiation of the `MediaTrack` item, means, the media player will show something old, as the new track has not been set yet. Neither are errors handled.

I've refactored the `MediaTrack` class to allow for a delayed setting of the source and some error handling if the token cannot be fetched. ~Currently, we report the song as ended, which would skip the song and continue to the next, which might fail again and thus continue to the end of the list.~ IMO, this is at least better than not handling errors. Now we have the choice to decide what we want to do. We could even refactor the `MediaTrack` to try setting the source again if we decide that that's a better strategy.

**UPDATE:** I've changed the code to pause if `getAccessTokenSilently()` fails to return a string. This way, the user could e.g. check his connection and resume at the song he wanted to listen to. After all, this is most likely not a problem which could be resolved by skipping a song, so this is IMO the better approach.

### DIFF
--- a/plugins/4.mediaPlayer.ts
+++ b/plugins/4.mediaPlayer.ts
@@ -2,6 +2,7 @@ import { useAuth0 } from "@auth0/auth0-vue";
 import { initMediaPlayer } from "./mediaPlayer/mediaPlayer";
 import type { AppInsights } from "./3.applicationInsights";
 import type { IUserData } from "./2.userData";
+import MediaTrack from "./mediaPlayer/MediaTrack";
 
 export default defineNuxtPlugin((_) => {
   const { getAccessTokenSilently } = useAuth0();
@@ -12,7 +13,7 @@ export default defineNuxtPlugin((_) => {
   return {
     provide: {
       mediaPlayer: initMediaPlayer(
-        (src) => new Audio(src),
+        () => new MediaTrack(),
         getAccessTokenSilently,
         appInsights,
         userData,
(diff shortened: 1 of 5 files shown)


plugins/4.mediaPlayer.ts
```diff
@@ -12,8 +13,10 @@ export default defineNuxtPlugin((_) => {
   return {
     provide: {
       mediaPlayer: initMediaPlayer(
-        (src) => new Audio(src),
-        getAccessTokenSilently,
+        (src) =>
+          new MediaTrack(() =>
+            getAccessTokenSilently().then((t) => authorizedUrl(src, t)),
```
kkuepper, line 18: The understanding from Andreas was that getAccessTokenSilently() is cached and that we don't need to optimize that.
Did you verify that that is not the case? Or is that an assumption (which might be wrong)?

SimonSimCity: I haven't changed anything on the workflow here. The reference has just been moved out of the class `MediaPlayer`.

The assumption, that `getAccessTokenSilently()` caches the token and only requests a fresh token when expired is true as long as `cacheMode` is set to `on`, which is the default - https://auth0.github.io/auth0-vue/interfaces/GetTokenSilentlyOptions.html#cacheMode.

kkuepper (approved):

Issue #337, opened 2024-03-02
Cannot read properties of undefined (reading 'json')


kboek:
When opening playlist "Daily Mix", the main window shows an error message `TypeError: Cannot read properties of undefined (reading 'json')`. The message doesn't go away even when you navigate to other items from the left pane. The only thing that helps is closing and re-opening the application

MacOS Sonoma 14.2.1
![Screenshot 2024-03-02 at 10 47 58](https://github.com/bcc-code/bmm-web/assets/743501/149a26d5-510c-43e4-922a-d8569014e288)

SimonSimCity:
Seems like the system failed while trying to report an error. This should be fixed in #343.

kkuepper:
fixed in v0.3.2

PR #343, opened 2024-03-09
Reporting all error responses

SimonSimCity:
but without requiring a change on the API. Furthermore, we can now better distinguish between two types of errors: 1. Errors most likely on the client (no response) and 2. an error response.


SimonSimCity:
This would actually fix #337.

kkuepper (approved):
lgtm

PR #344, opened 2024-03-10
Reset audio on error

### LINKED ISSUE #193: Player: 403 error when trying to resume player after long break (Signed Cloudfront url expires)
The network request for the mp3 failed after trying to resume after leaving the website tab in the background.
I forgot to log more details like the access token etc before I closed the tab. If I can make it happen again I'll post more details.
Here's all I have.
Chromium-based edge on mac.

**Repro steps as far as I know:**
1. Play something on web (I used chromium-based edge)
2. Pause the thing and leave it in the background for a while
3. Come back a long time later (the tab was also [sleeping](https://www.microsoft.com/en-us/edge/features/sleeping-tabs-at-work)
4. Try resuming

**Error**
somethingsomething.mp3
```console
Failed to load resource: the server responded with a status of 403 ()
```

SimonSimCity:
This PR includes the changes of #342 and adds 3 commits in an attempt to fix #193.

The audio-element we use as an error-event which [_[...] is fired when the resource could not be loaded due to an error (for example, a network connectivity problem)._](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/error_event)

In this case, an instance of `MediaError` is available at the `error` property, which has two properties: `message` whose content is not specified in the specs and could therefore be anything, and `code` - an integer as further described below.

Name | Value | Description
-- | -- | --
MEDIA_ERR_ABORTED | 1 | The fetching of the associated resource was aborted by the user's request.
MEDIA_ERR_NETWORK | 2 | Some kind of network error occurred which prevented the media from being successfully fetched, despite having previously been available.
MEDIA_ERR_DECODE | 3 | Despite having previously been determined to be usable, an error occurred while trying to decode the media resource, resulting in an error.
MEDIA_ERR_SRC_NOT_SUPPORTED | 4 | The associated resource or media provider object (such as a [MediaStream](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream)) has been found to be unsuitable.

When disconnecting from the network I got both `2` and `4`, which makes it unreliable to me which code the user will get for what error. For this reason I chose to reinitialize the track (the error could be an expired token or any other reason) and pause it, so the user can resume it. This is the best we can do with the tools we have at hand at the moment.

To have a clearer set of changes, this PR should be merged after being rebased after merging #342.


plugins/mediaPlayer/MediaTrack.ts
```diff
@@ -163,10 +192,34 @@ export default class MediaTrack {
...
+      */
+      // On error, reset the audio instance.
+      this.audioElement = new Audio();
+      this.audioElement.autoplay = true;
+      this.registerEvents();
+      // Set the time to where the error was thrown, so the user can continue.
+      this.position = this.p;
+      // Pause so the user has to trigger an action. If we init the source, we might run into an infinite loop here.
+      this.paused = true;
+      this.loading = false;
```
kkuepper, line 213: can you please add a log to appInsights

SimonSimCity: That's done in line 196, which calls the method set in `plugins/4.mediaPlayer.ts` as second argument when instantiating `MediaTrack`. This way we have it decoupled from the actual implementation which makes for easier testing.

SimonSimCity:
I'll have to check the linter. Seems like I've overlooked something...

PR #346, opened 2024-03-11
hide titlebar & don't show logo on desktop apps

kkuepper:


components/sidebar/SidebarElement.vue
```diff
@@ -2,15 +2,24 @@
 import { version } from "~/package.json";
 
 const { data: collections } = usePrivatePlaylists();
+const isElectron = window.isElectron || false;
```
SimonSimCity, line 5: What about using `runtimeConfig.public.systemName` instead of introducing yet another variable..?

components/sidebar/SidebarElement.vue
```diff
@@ -2,15 +2,24 @@
...
     class="flex-none flex max-h-screen w-[300px] flex-col border-r border-label-separator bg-background-2"
   >
-    <div class="flex items-center p-3">
-      <SiteLogo size="small" />
+    <div
+      class="flex items-center p-3"
+      style="
+        -webkit-app-region: drag;
+        -webkit-user-select: none;
+        user-select: none;
```
SimonSimCity, line 17: Why have those 3 CSS options been added to a couple of places? Shouldn't `user-select: none` be applied to the all the page? And what is it about `-webkit-app-region: drag`?

components/ProfileMenu.vue
```diff
@@ -192,21 +193,34 @@ const saveAndCloseContentLanguageDialog = () => {
...
                 }"
-                class="flex w-full rounded-lg px-3 py-2"
+                class="flex w-full rounded-lg px-3 py-2 font-medium text-[15px]"
                 href="mailto:bmm-support@bcc.no"
                 target="_blank"
               >
                 {{ $t("profile.contact") }}
               </a>
             </MenuItem>
+            <MenuItem v-slot="{ active }" as="div">
```
SimonSimCity, line 203: Won't the version be shown two times on Web this way..?

components/sidebar/SidebarElement.vue
kkuepper: Currently, you can drag the window using the titlebar. When the titlebar is hidden you have no way of moving the window. Therefore I defined the AppToolbar and the logo area as draggable so that you can move the window.

It might be a good idea to apply user-select: none everywhere, but I'm not sure if that messes with anything. Also, we probably shouldn't do it on the website.

components/ProfileMenu.vue
kkuepper: Yes. Maybe it's better to deviate from Søren's design and add the logo on the desktop app.

components/sidebar/SidebarElement.vue
kkuepper: Didn't know about that one. Sounds like a better idea.

components/sidebar/SidebarElement.vue
kkuepper: I'm now using user-select for the whole app if it's on Electron.

components/ProfileMenu.vue
kkuepper: re-added the logo and removed the profile entry

PR #347, opened 2024-03-13
implement design for recommendations

kkuepper:


components/DocumentList.vue
```diff
@@ -48,9 +67,7 @@ const convertModels = (models: IAllDocumentModels[]) => {
       currentSection.push(el);
     }
   });
-  if (tiles.length > 0) {
-    result.splice(1, 0, { header: null, items: tiles, useFlex: true });
-  }
+  console.log("models", result);
```
sifferhans, line 70: This should probably be removed

plugins/1.auth0.client.ts
```diff
@@ -9,7 +9,7 @@ export default defineNuxtPlugin((nuxtApp) => {
     cacheLocation: "localstorage",
     authorizationParams: {
       redirect_uri: window.location.origin,
-      audience: runtimeConfig.public.apiUrl,
+      audience: "https://bmm-api.brunstad.org",
```
sifferhans, line 12: What is the reason for this? Isn't it ok to keep it in the runtime config? I guess the url for the API doesn't change

sifferhans (approved):

plugins/1.auth0.client.ts
kkuepper: auth0 should always use bmm-api even when the apiUrl changes to int-bmm-api.brunstad.org.

I used int-bmm-api for this PR since on INT we show an example of all possible recommendations.

components/DocumentList.vue
kkuepper: thanks

components/RecommendationItem.vue
```diff
@@ -0,0 +1,69 @@
...
+const { t } = useI18n();
+
+defineProps<{
+  item: RecommendationModel;
+}>();
+</script>
+
+<template>
+  <div class="col-span-full bg-background-2 my-2 rounded-2xl border-t-0">
+    {{ console.log("item", item) }}
```
SimonSimCity, line 13: This should've been removed 😅

SimonSimCity (commented):
Looks good otherwise.

PR #350, opened 2024-03-13
add styling for subalbums

kkuepper:


sifferhans (approved):
Looks good to me 👍🏼

components/AlbumDetails.vue
```diff
@@ -41,34 +51,26 @@ const toggleExpandedAlbum = (albumReference: string) => {
...
-            </ButtonStyled>
+            <ButtonStyled icon="icon.link" icon-only></ButtonStyled>
           </CopyToClipboard>
         </div>
       </div>
     </header>
-    <p v-if="album.children" class="p-2">
-      {{ t("collection.album-count", album.children.length) }}
+    <p v-if="album.children" class="py-2 text-label-3">
+      <template v-if="childAlbums.length > 0">
```
SimonSimCity, line 60: What if it has both? What if an album contains both albums and tracks as children? We have that at least once.

components/album/SubAlbum.vue
```diff
@@ -14,40 +15,70 @@ const emit = defineEmits<{ expand: [] }>();
...
-        {{ t("collection.track-count", album?.children.length) }}
-      </p>
+    <section v-if="album" class="relative flex items-center gap-4">
+      <CoverImage :src="album.cover" class="w-[68px] rounded-md" />
+      <div class="text-2xl font-extrabold">{{ album.title }}</div>
+      <div
+        v-if="album?.children && !active"
+        class="flex gap-1.5 ml-auto px-4 py-2 bg-background-2 rounded-3xl border-[1px] border-background-2 group-hover:border-label-separator"
+      >
+        <span class="font-semibold text-lg leading-6">{{
```
SimonSimCity, line 44: Do we currently only allow for albums containing sub-albums or do we have deeper nested structure? I guess we agreed on that we won't. In this case this code is fine. Otherwise we'd have to improve here.

SimonSimCity (commented):
Looks good otherwise.

components/album/SubAlbum.vue
SimonSimCity: I guess this discussion has been confirmed earlier, we have just one level of nesting here.

components/AlbumDetails.vue
SimonSimCity: Will be fixed in an upcoming PR. There's another album, "Påskecamp 2024", where we again have an album containing tracks and albums.

PR #353, opened 2024-03-15
reuse same shadow for player

kkuepper:


components/album/SubAlbum.vue
```diff
@@ -92,15 +92,15 @@ const childTracks = computed(() => {
     </section>
   </section>
 </template>
-<style scoped>
-.album-shadow {
+<style>
```
SimonSimCity, line 95: Non-scoped styles should rather go into a general CSS file. It's quite tempting, but it seems better if something which yield global also is global.

If we e.g. would refactor this file, we might not remember that those CSS things are used by a different component. In my option, this makes things quite complex and interdependent.

PR #354, opened 2024-03-15
Feature/290 endless scroll

kkuepper:


components/EndlessDocumentList.vue
```diff
@@ -0,0 +1,65 @@
...
+          console.error("error", e);
+        });
+    },
+    { distance: 10, interval: 500, canLoadMore: () => !fullyLoaded.value },
+  );
+});
+watch(
+  [useNuxtApp().$i18n.locale, () => contentLanguageStore().contentLanguages],
+  async () => {
+    const data = await props.load(0, position);
```
SimonSimCity, line 47: We should also handle errors here... You do it well above, but here you leave errors to the watch-implementation to deal with.

components/EndlessDocumentList.vue
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      loadingMore.value = true;
+      props
+        .load(position, 40)
+        .then((data) => {
+          position += 40;
+          if (data) {
+            if (data.length === 0) {
+              console.log(`list is fully loaded. position: ${position}`);
```
SimonSimCity, line 29: Should we leave this console-log in?

SimonSimCity (changes requested):
Looks good to me otherwise.

components/EndlessDocumentList.vue
kkuepper: we're just swallowing the exception.
If it happens often, we should invest some time to figure out what's the best user experience.

PR #362, opened 2024-03-26
Small adjustments

SimonSimCity:
Just some things I saw could do good of some improvements in the last PRs.

Why I implemented [Dynamic width for context menu](https://github.com/bcc-code/bmm-web/commit/154599d5253011f9aeb0c3869ec07d46339cf372) is best explained in these two screenshots:

**Before:**
before
**After:**
after

### DIFF
--- a/pages/playlist/contributor/[id]/index.vue
+++ b/pages/playlist/contributor/[id]/index.vue
@@ -21,11 +21,16 @@ const onPressPlay = () => {
   }
 };
 const onPressShuffle = async () => {
-  const shuffledTracks = (await useContributorShuffle(contributorId)).data
-    .value;
+  try {
+    const shuffledTracks = (await useContributorShuffle(contributorId)).data
+      .value;
 
-  if (shuffledTracks) {
-    setQueue(shuffledTracks);
+    if (shuffledTracks) {
+      setQueue(shuffledTracks);
+    }
+  } catch (e) {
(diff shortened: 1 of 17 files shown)


pages/playlist/contributor/[id]/index.vue
```diff
@@ -21,11 +21,16 @@ const onPressPlay = () => {
...
+  try {
+    const shuffledTracks = (await useContributorShuffle(contributorId)).data
+      .value;
 
-  if (shuffledTracks) {
-    setQueue(shuffledTracks);
+    if (shuffledTracks) {
+      setQueue(shuffledTracks);
+    }
+  } catch (e) {
```
Comment (kkuepper) [inline pages/playlist/contributor/[id]/index.vue:31]: I don't like that every api call needs to be wrapped in a try catch. I'd prefer if we detect the error in the middleware and show a message. Then not every component needs to do it itself.

kkuepper (approved):

pages/playlist/contributor/[id]/index.vue
Comment (SimonSimCity) [reply pages/playlist/contributor/[id]/index.vue:31]: That's part of #33. I wouldn't do this in a middleare, because some errors are expected to be handled, others are rather general. I'd rather wrap them all in a wrapper where the caller itself can handle the expected ones and let the wrapper catch all other errors, which should also be reported differently, IMO. That's also why I'm rather against the current implementation of error reporting.

PR #366, opened 2024-03-27
Added the track-repeat mode

SimonSimCity:
... and added some tests and refactored the changes of #319.

Please read the commit notes for further details.

### DIFF
--- a/plugins/mediaPlayer/mediaPlayer.ts
+++ b/plugins/mediaPlayer/mediaPlayer.ts
@@ -15,6 +15,7 @@ export enum MediaPlayerStatus {
 export enum RepeatStatus {
   NoRepeat = 0,
   RepeatQueue = 1,
+  RepeatTrack = 2,
 }
 
 export interface MediaPlayer {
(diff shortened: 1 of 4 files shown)


plugins/mediaPlayer/mediaPlayer.ts
```diff
@@ -87,6 +89,29 @@ export const initMediaPlayer = (
     activeMedia.value.registerEvents();
   }
 
+  function restartTrack() {
+    // TODO: Can we somehow reset the media-track?
+    initCurrentTrack();
+    // activeMedia.value!.ended = false;
```
kkuepper, line 95: looks like some leftover

SimonSimCity: That's a try of the previous todo. Should I mark it a bit different? It's a code-basis where implementing the open task could start from.

The open task is to avoid reinitializing the track, which requires another download of the file, which shouldn't be necessary. I haven't tested resetting the ended flag and the position. This could work already...

kkuepper: right now it looks like some leftover code and it's not clear why it's there. Either put the content of your reply into the code so that it's more understandable, or remove the commented code.

SimonSimCity: True. Please see my changed code. This should give an idea of what I wanted to bring across. It shouldn't be understood as a ready-to-go code, or as a comment because I wanted to remove something, but rather as a start of how to fix the mentioned task.

kkuepper (approved):

PR #373, opened 2024-03-30
Page title

### LINKED ISSUE #260: Title of the website
There was already a small start by @sifferhans in the PR #258, which I asked him to roll back.

This ticket is first and foremost to create a concept of what should be the title of the website at any given time. @sclausendk, I think you should have a good first start in this discussion.

At the moment, the title is fixed to `BMM`.

SimonSimCity:
Fixes #260.
Based on #372.

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -102,7 +102,7 @@ useProfileStore().$subscribe((_, state) => {
 });
 
 useHead({
-  titleTemplate: (chunk) => (chunk ? `BMM | ${chunk}` : "BMM"),
+  titleTemplate: (chunk) => (chunk ? `BMM - ${chunk}` : "BMM"),
 });
 </script>
 
--- a/app.vue
+++ b/app.vue
@@ -102,7 +102,7 @@ useProfileStore().$subscribe((_, state) => {
 });
 
 useHead({
-  titleTemplate: (chunk) => (chunk ? `BMM - ${chunk}` : "BMM"),
+  titleTemplate: (chunk) => (chunk ? `${chunk} - BMM` : "BMM"),
 });
 </script>
 
--- a/locales/en.json
+++ b/locales/en.json
@@ -15,7 +15,7 @@
     "search": "Search",
     "album": "Album",
     "podcast": "Podcast",
-    "unknown": "Unknown route",
+    "not-found": "Page not found",
     "contributor": "Contributor",
     "contributors": "Contributors",
     "archive": "Archive"
--- a/locales/en.json
+++ b/locales/en.json
@@ -15,6 +15,7 @@
     "search": "Search",
     "album": "Album",
     "podcast": "Podcast",
+    "unknown": "Unknown route",
     "contributor": "Contributor",
     "contributors": "Contributors",
     "archive": "Archive"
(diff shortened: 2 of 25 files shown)


pages/[...error].vue
```diff
@@ -1,5 +1,10 @@
 <script setup lang="ts">
+const { t } = useI18n();
 const { params } = useRoute<"error">();
+toolbarTitleStore().setReactiveToolbarTitle(() => t("nav.unknown"));
```
Comment (kkuepper) [inline pages/[...error].vue:4]: not sure if it's a good idea or not. But maybe we could just use the Toolbar title if no useHead.title is provided.
That would simplify a lot of pages

Comment (SimonSimCity) [reply pages/[...error].vue:4]: True.. my first thought was that this wouldn't be in line with what I agreed on with Søren, but when I now think it through, it seems also hard to realize without having some glitches here and there 😅 I think we'll have to define it at the initialization of every page, and I think this makes you also aware of that there are two titles which you have to decide how you want them to be.

SimonSimCity:
@kkuepper How about this? I'd merge it as is and let the programmer specifically decide what should be in the title and what should be in the toolbar. To me, this is a decision which should be taken for each side and not "by accident", which is why I'd keep it this way to set it for each page.

kkuepper:
@SimonSimCity  can we wait two weeks until I’m back in the office? Then I’ll propose some code  to improve it.

pages/[...error].vue
Comment (kkuepper) [reply pages/[...error].vue]: @SimonSimCity Can you have a look at my commit and give feedback on it?

app.vue
```diff
@@ -102,7 +102,7 @@ useProfileStore().$subscribe((_, state) => {
 });
 
 useHead({
-  titleTemplate: (chunk) => (chunk ? `BMM | ${chunk}` : "BMM"),
+  titleTemplate: (chunk) => (chunk ? `${chunk} - BMM` : "BMM"),
```
kkuepper, line 105: I think at some point I suggested putting BMM first. But after looking at other websites (Figma, Google, Microsoft Azure, ...) I realized it's commonly done the other way.

locales/en.json
```diff
@@ -15,6 +15,7 @@
     "search": "Search",
     "album": "Album",
     "podcast": "Podcast",
+    "unknown": "Unknown route",
```
sifferhans, line 18: Should this say "Unknown page" instead? I don't know if "route" makes sense for end users

pages/search/[[term]].vue
```diff
@@ -10,6 +10,9 @@ import type {
 const router = useRouter();
 const { t } = useI18n();
 toolbarTitleStore().setReactiveToolbarTitle(() => "");
```
Comment (sifferhans) [inline pages/search/[[term]].vue:12]: This should probably we replaced with the `setTitle` function you introduced in other pages

Comment (sifferhans) [reply pages/search/[[term]].vue:12]: Ah, nevermind. I see that the `setTitle` function only should be used if the toolbar title and page title are the same

sifferhans (changes requested):
Just the "Unknown page" translation

locales/en.json
kkuepper: I agree. Change it to "Page not found"

sifferhans (approved):

Issue #288, opened 2024-02-09
Download files


kkuepper:
- [ ] implement download functionality (url is on track.media.files.url. Take first of array)
- [ ] show popup when trying to download a message ([see figma](https://www.figma.com/file/vUnfD52FdhTmXq4Wjd5ieI/Bmm-Web-2023?type=design&node-id=737-18349&mode=design&t=DbpiV6qbBweK6DUC-0))
- [ ] don't show popup if user has Download permission
- [ ] hide download button on Electron

This isn't strictly relevant for Public release. We could just hide the download button.

The text in Figma is outdated:
headline: Download of messages unavailable
text: The ability to download messages from BMM is no longer available. If you need a message for a specific use, please contact us and explain your needs.
support mail: mailto:support@bcc.media

kkuepper:
I decided against implementing the download action, since we're considering removing it at some point anyway. Because of that it shouldn't be one of the quick actions. So I use "AddToPlaylist" instead as quick action.
Also downloading in Electron doesn't really feel right. We should still implement it for the website though.

PR #382, opened 2024-04-15
feat: download files on web #288

andreasgangso:


utils/tracks.ts
```diff
@@ -0,0 +1,18 @@
...
+  TrackModelMediumFile,
+} from "@bcc-code/bmm-sdk-fetch";
+
+export function defaultFileInMedia(
+  media: TrackModelMedium[],
+): TrackModelMediumFile | undefined {
+  return media.find((m) => m.files != null)?.files![0];
+}
+
+export function defaultFileForTrack(
```
kkuepper, line 13: +1 for extracting that

components/track/TrackMenu.vue
```diff
@@ -47,6 +51,19 @@ const dropdownMenuItemsForTrack = (track: TrackModel) => {
     });
   }
 
+  if (runtimeConfig.public.systemName !== "Electron") {
+    items.push({
+      icon: "icon.download",
+      text: t("track.dropdown.download"),
+      clickFunction: async () => {
+        const result = await download(track);
+        if (result === "no-permission") {
```
kkuepper, line 60: is there a better way than having "magic strings"?
For example by splitting it into to parts:
- hasPermissionToDownload()
- download()

Another way would be to introduce a strongly typed return value

components/dialog/DialogDownloadNotAllowed.vue
```diff
@@ -0,0 +1,44 @@
...
+        {{ $t("download.messages-not-allowed.description") }}
+      </p>
+      <TooltipBase placement="top" :offset="8" class="mb-3 mt-5">
+        <ButtonStyled intent="primary" @click="contact">
+          {{ $t("profile.contact") }}
+        </ButtonStyled>
+        <template #content>
+          <div class="rounded-2xl bg-background-1 px-5 py-3 shadow-md">
+            <p class="text-sm text-label-2">
+              {{ $t("download.sendEmailTo", ["support@bcc.media"]) }}
```
kkuepper, line 37: can you extract the email into a variable please?

kkuepper (changes requested):

components/track/TrackMenu.vue
andreasgangso: is **is** a strongly typed return value: DownloadResult.

components/dialog/DialogDownloadNotAllowed.vue
andreasgangso: idk exactly in what form you thought, so I went to the other extreme and made an environment variable :)

components/dialog/DialogDownloadNotAllowed.vue
```diff
@@ -0,0 +1,46 @@
...
+      <NuxtIcon name="download" class="mt-5 text-center text-5xl" />
+      <h3 class="my-3 text-2xl font-extrabold leading-tight text-label-1">
+        {{ $t("download.messages-not-allowed.title") }}
+      </h3>
+
+      <p class="px-10 text-label-2">
+        {{ $t("download.messages-not-allowed.description") }}
+      </p>
+      <TooltipBase placement="top" :offset="8" class="mb-3 mt-5">
+        <ButtonStyled intent="primary" @click="contact">
```
sifferhans, line 33: Isn't it better to use an `a` tag here, since it basically is a navigation?

utils/download.ts
```diff
@@ -0,0 +1,29 @@
...
+  const file = defaultFileForTrack(track);
+  const url = file?.url;
+  if (file && url) {
+    const authUrl = authorizedUrl(url, accessToken);
+    window.location.href = `${authUrl}&download=true`;
+  }
+}
+
+export type DownloadResult = "no-permission" | "downloaded";
+export function useTrackDownloader() {
```
sifferhans, line 14: This is a composable, and should be in the `composables` folder, strictly speaking. 
It is quite practical to keep everything in one place, but I think we should try following Nuxt conventions.

sifferhans (changes requested):
We should move the `useTrackDownloader` composable to the `composables` directory, but other than that it looks good to me :)

components/dialog/DialogDownloadNotAllowed.vue
andreasgangso: true but i'd rather use <a> for simplicity

sifferhans (approved):

kkuepper (approved):

PR #384, opened 2024-04-15
refactor: generate typography tokens

andreasgangso:
- Removes generation of tailwindconfig so that you can customize it without modifying the generator scripts.
- Generates `type-` classes from the tokens.json, which should 1-1 match figma. These should show up with [intellisense](https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss).
image
- Removes gitignore for generated files. Ignoring files is good when the files are only used in development, or are very environment-specific. But these files are simple and constant and used at runtime, so it's helpful to version-control them.


nuxt.config.ts
```diff
@@ -48,7 +48,22 @@ export default defineNuxtConfig({
...
+
+      link: [
+        { href: "https://fonts.googleapis.com", rel: "preconnect" },
+        {
+          href: "https://fonts.gstatic.com",
+          rel: "preconnect",
+          crossorigin: "anonymous",
+        },
+        {
+          href: "https://fonts.googleapis.com/css2?family=Inter:wght@500;600;700;800&display=swap",
```
kkuepper, line 62: I'm pretty sure we use weight 400 somewhere. Maybe 900 as well.

assets/design-tokens/tailwind-colors.generated.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * ---------------------------------------------------
+ * 👷 Generated by figma2tailwind. Do not modify !
```
kkuepper, line 3: why is this file committed if it's generated?

assets/design-tokens/colors.generated.css
kkuepper, line 1: why is this file committed if it's generated?

.prettierignore
```diff
@@ -13,4 +13,4 @@ pnpm-lock.yaml
 public/staticwebapp.config.json
 
 # Generated files
-tailwind.config.ts
+*.generated.*
```
kkuepper, line 16: I wonder if this isn't working as expected

nuxt.config.ts
andreasgangso: The design system doesn't include those, so that's kinda wrong. but sure, no biggie to include those. done.

assets/design-tokens/tailwind-colors.generated.ts
andreasgangso: You didn't read my PR comment? x)

.prettierignore
andreasgangso: why? did you confuse this with gitignore?

kkuepper: yup. totally confused it with gitignore

assets/design-tokens/tailwind-colors.generated.ts
kkuepper: comment you mentioned: "Removes gitignore for generated files. Ignoring files is good when the files are only used in development, or are very environment-specific. But these files are simple and constant and used at runtime, so it's helpful to version-control them."

I'm not sure if I agree with that. Let's talk about it tomorrow

nuxt.config.ts
kkuepper: I was wrong. They aren't used

PR #388, opened 2024-04-16
fix: prevent floating-ui from messing up dialogs

andreasgangso:
dunno why, but the dialog transitions doesnt like floating-ui.
the dialog transitions on the download dialog are gone (it just snaps in), unless we delay the tooltip by 1 tick x)


kkuepper:
I actually can't test it since I have the download permission.
@sifferhans or @SimonSimCity can you test it?

andreasgangso:
I tested it myself 👍

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

kkuepper (approved):

PR #393, opened 2024-04-18
Feature/track played event

kkuepper:


plugins/mediaPlayer/MediaTrack.spec.ts
```diff
@@ -15,13 +16,19 @@ class MediaTrackMock extends MediaTrack {
   }
 }
 
+const appInsights = {
+  event: (_: string, _2: any) => {},
+} as unknown as AppInsights;
```
sifferhans, line 21: What is the reason for casting to `unknown` and `AppInsights` here? Is it not possible to do as such?:
```ts
const appInsights: AppInsights = {
  event: (_: string, _2: any) => {},
}
```

I realize this is just a test file, so it isn't that critical, but I am just curious about the reasoning here.

kkuepper: nice catch. I just copied it without thinking about it.

PR #395, opened 2024-04-19
more helpful error page

kkuepper:


package.json
kkuepper, line 1: Those updates aren't strictly necessary, but I also don't see a reason to reverse them.

PR #412, opened 2024-05-02
feat: floating ui for dropdowns

andreasgangso:


.eslintrc.json
```diff
@@ -100,12 +100,13 @@
...
-    "@intlify/vue-i18n/no-missing-keys-in-other-locales": "warn",
-    "@intlify/vue-i18n/no-dynamic-keys": "warn",
+    "@intlify/vue-i18n/no-raw-text": "off",
+    "@intlify/vue-i18n/no-dynamic-keys": "off",
     "@intlify/vue-i18n/no-unused-keys": [
       "warn",
       { "extensions": [".vue", ".ts"] }
-    ]
+    ],
+    "vue/no-v-html": "off"
```
andreasgangso, line 109: couldnt get this to work.. i dont think v-html is a necessary lint since we rarely show user input anyway

kkuepper:
functionality looks good. I'll review the code tomorrow

.eslintrc.json
SimonSimCity, line 1: I vote for reverting the changes here. I'm all for keeping warnings low, but I wouldn't do it for the expense of making unexperienced programmers unaware of these things.

components/dropdown/DropdownMenu.vue
```diff
@@ -0,0 +1,101 @@
...
+  MenuItems,
+  TransitionRoot,
+  TransitionChild,
+} from "@headlessui/vue";
+
+const props = defineProps<{
+  placement?: Placement;
+}>();
+
+const buttonRef = ref<null | any>(null);
```
SimonSimCity, line 23: Can we avoid using `any` here?

components/track/TrackItem.vue
```diff
@@ -69,6 +67,7 @@ const selectedTrack: Ref<TrackModel | null> = ref(null);
       class="absolute -inset-x-4 -inset-y-0 rounded-xl bg-tint"
     ></div>
 
+    <!-- eslint-disable vue/no-v-html -->
```
SimonSimCity, line 70: Is disabling the eslint-rule here needed ..?

modules/icons/runtime/components/nuxt-icon.vue
```diff
@@ -30,7 +30,7 @@ async function getIcon() {
   }
 }
 
-await getIcon();
+getIcon();
```
SimonSimCity, line 33: `getIcon` is an async function and should be called with an await, or the possible error should be handled otherwise. Leaving it like this is negligent IMO.

tsconfig.json
```diff
@@ -7,6 +7,6 @@
     "lib": ["ESNext", "DOM"],
     "noUncheckedIndexedAccess": true,
     "verbatimModuleSyntax": true, // Can be removed after updating nuxt to v3.8
-    "exactOptionalPropertyTypes": true
+    "exactOptionalPropertyTypes": false // disabled because of https://github.com/vuejs/core/issues/6532
```
SimonSimCity, line 10: What's bugging me is why we haven't had this issue before ... Do you know what changes are causing this issue?

SimonSimCity (changes requested):

components/track/TrackItem.vue
andreasgangso: wasnt able to turn it off in eslintrc 🤷‍♂️ tried for a while actually

tsconfig.json
andreasgangso: no 🤷

modules/icons/runtime/components/nuxt-icon.vue
andreasgangso: Nuxt handles the error but it's a nasty fullscreen error. Moved the try/catch and replaced it with console.warn instead of throwing 👍 

But awaiting in `setup` is not fun, as the component becomes an async component, and you start needing <suspense> and stuff.

Awaiting here causes the icon component to not be rendered the first few frames, and that causes layout shifting when it finally appears (especially noticable inside dropdowns). It's better to not await and render a placeholder box the same size as the icon instead.

components/dropdown/DropdownMenu.vue
andreasgangso: yea I gave up trying to find the correct type 💀 but found it now

.eslintrc.json
andreasgangso: eslint by default at least makes these errors very annoying. when they aren't fixable it's annoying and you'll just become blind on them. It can also be very time consuming to tweak.

- no-dynamic-keys: reverted. there are cases where this is useful, but it's maybe only 1 place in the codebase so I replaced with `<!-- eslint-disable-next-line -->`
- v-html: It's very rarely if ever relevant for us to have user input here. And it's not fixable. So it would just annoy/confuse beginners. We have code reviews.
- vue-i18n/no-missing-keys-in-other-locales: removed by mistake. added back now.
- no-raw-text: It's too aggressive. Stuff like [this](https://github.com/bcc-code/bmm-web/commit/9d82a700baa6b794288eeb38fdfb0fb364443354) causes warnings because " (" is raw text.. One could work on a regex to improve this but probably overkill.

components/PrivatePlaylistMenu.vue
```diff
@@ -1,19 +1,12 @@
...
 import { TrackCollectionApi } from "@bcc-code/bmm-sdk-fetch";
 import type { GetTrackCollectionModel } from "@bcc-code/bmm-sdk-fetch";
-import { Menu, MenuButton, MenuItem, MenuItems } from "@headlessui/vue";
 
 const { t } = useI18n();
 
-const props = withDefaults(
-  defineProps<{
-    playlist: GetTrackCollectionModel;
-    buttonClass?: string;
```
kkuepper, line 11: don't we need this?

components/PrivatePlaylistMenu.vue
```diff
@@ -88,55 +81,30 @@ const dropdownMenuItems = () => {
...
-    </MenuItems>
-  </Menu>
+        </DropdownMenuItem>
+      </DropdownMenuGroup>
+    </template>
+  </DropdownMenu>
   <DialogBase
     :show="showEditDialog"
-    :title="$t('playlist.new-playlist')"
+    :title="$t('playlist.name-your-playlist')"
```
kkuepper, line 107: this looks suspicious. Was it a bug before?

components/track/TrackMenu.vue
```diff
@@ -116,52 +114,29 @@ const dropdownMenuItemsForTrack = (track: TrackModel) => {
...
 <template>
-  <Menu
-    as="div"
-    class="relative flex flex-col justify-center text-left"
-    :class="$attrs.class"
-    @click.stop
-  >
-    <MenuButton
-      as="button"
+  <DropdownMenu placement="bottom-end" v-bind="$attrs" @click.stop>
```
kkuepper, line 117: I like that all cases now require less code to achieve the same. Well done.

components/PrivatePlaylistMenu.vue
kkuepper: I guess it was copied from TrackMenu without being needed.

components/PrivatePlaylistMenu.vue
kkuepper: yup. was a bug

kkuepper (approved):

PR #419, opened 2024-05-09
 copy solution for signing Windows .exe from bcc-code/bible-explorers

kkuepper:


cs1m0n (commented):
What you could do is merge the two specifically-for-Windows steps together (**Install AzureSignTool** and **Package the Electron app**).
Otherwise, LGTM!

kkuepper:
@cs1m0n I think it's more readable as separate steps.

PR #421, opened 2024-05-14
Volume control & draggable sliders

kkuepper:


andreasgangso (changes requested):
didnt read detailed but id refactor into a reusable slider component instead of embedding it inside the mediaplayer.

also consider https://zagjs.com/components/vue-sfc/slider for some free functionality and accessibility (zag is modular, pnpm install this and import)

kkuepper:
I looked at different sliders but didn't like them for different reasons. But the slider from zag looks really good. I'll refactor to use it.

components/media-player/MediaPlayerOpen.vue
```diff
@@ -24,22 +26,68 @@ const {
...
+  volume,
 } = useNuxtApp().$mediaPlayer;
 
-const onPointerUpProgressBar = (event: PointerEvent) => {
-  const rect = (event.currentTarget as Element)?.getBoundingClientRect();
-  currentPosition.value =
-    ((event.clientX - rect.left) / rect.width) * currentTrackDuration.value;
+const [state, send] = useMachine(
+  slider.machine({
+    id: "volume",
```
kkuepper, line 34: @andreasgangso I refactored the volume slider. Do you by chance know how to make a 2nd slider work?
I haven't really tried it much but my first try caused some problems.

kkuepper: done

components/media-player/PositionSlider.vue
```diff
@@ -0,0 +1,90 @@
...
+
+const currentOrNewPosition = computed(() =>
+  newPosition.value === null
+    ? currentPosition.value
+    : (newPosition.value / 100) * currentTrackDuration.value,
+);
+
+watch([currentPosition, currentTrackDuration], () => {
+  if (newPosition.value === null) {
+    const value = (currentPosition.value / currentTrackDuration.value) * 100;
```
andreasgangso, line 38: is division by zero possible?

plugins/mediaPlayer/mediaPlayer.ts
```diff
@@ -193,6 +197,16 @@ export const initMediaPlayer = (
     },
   });
 
+  const volumeComputed = computed({
```
andreasgangso, line 200: id prefer exposing volume and doing:
```ts
watch(volume, (v) => {
  activeMedia.value?.setVolume(v);
});
```

components/media-player/PositionSlider.vue
```diff
@@ -0,0 +1,90 @@
...
+    id: "position",
+    value: [0],
+    step: 0.01,
+    onValueChange(details) {
+      const [value] = details.value;
+      if (value) newPosition.value = value;
+    },
+    onValueChangeEnd(details) {
+      const [value] = details.value;
+      if (value) {
```
andreasgangso, line 19: why the if (value)? is it nullable?
0 is falsy, so maybe you want to do `value != null` (null or undefined) instead?

components/media-player/VolumeSlider.vue
```diff
@@ -0,0 +1,69 @@
...
+
+const { volume } = useNuxtApp().$mediaPlayer;
+
+const [volumeState, volumeSend] = useMachine(
+  slider.machine({
+    id: "volume",
+    value: [volume.value * 100],
+    onValueChange(details) {
+      const [value] = details.value;
+      if (value) volume.value = value / 100;
```
andreasgangso, line 13: if (value) ?

andreasgangso (changes requested):
some comments

plugins/mediaPlayer/mediaPlayer.ts
kkuepper: I think the current solution is more in line with what is already being done in the mediaPlayer

components/media-player/PositionSlider.vue
kkuepper: zag-js defines the value as number | undefined

components/media-player/PositionSlider.vue
kkuepper: changed it

PR #424, opened 2024-05-27
Feature/improve showing songs

kkuepper:
I still need to show the song number in the player

### DIFF
--- a/components/track/TrackItem.vue
+++ b/components/track/TrackItem.vue
@@ -48,6 +48,16 @@ function adjustHighlightText(highlight: Highlighting) {
     .replaceAll("**/", "</span>");
 }
 
+function getSongNumber(track: TrackModel) {
+  if (track.songbookRelations && track.songbookRelations.length > 0) {
+    const rel = track.songbookRelations[0];
+    if (rel) {
+      return `${songbookName(rel.name)} ${rel.id}`;
+    }
+  }
+  return null;
+}
+
 const isPlaying = computed(() => currentTrack.value?.id === props.track.id);
 
 const selectedTrack: Ref<TrackModel | null> = ref(null);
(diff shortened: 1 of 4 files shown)


components/track/TrackItem.vue
```diff
@@ -48,6 +48,33 @@ function adjustHighlightText(highlight: Highlighting) {
...
 }
 
+const fields = computed(() => {
+  const parts = [
+    props.track.title,
+    trackSongNumber(props.track),
+    props.track.meta.artist,
+    props.isTrackTypeKnown ? null : props.track.meta.album,
+  ];
+  const filtered = parts.filter((part) => part);
```
sifferhans, line 58: We could use `parts.filter(Boolean)` here as well, but it is basically the same

components/track/TrackItem.vue
```diff
@@ -48,6 +48,33 @@ function adjustHighlightText(highlight: Highlighting) {
     .replaceAll("**/", "</span>");
 }
 
+const fields = computed(() => {
```
sifferhans, line 51: Some of this logic is repeated in the `trackFields` function in `utils/trackFields.ts`.
Could we use the util here, or is the logic too custom here?

sifferhans (approved):
Only thing is the repeated logic in `TrackItem.vue`

components/track/TrackItem.vue
kkuepper: TrackItem has some special logic. I thought it makes sense to keep the 2 separate.
But I'm happy the logic now is only in 2 places instead of 5.

components/track/TrackItem.vue
kkuepper: I don't understand your suggestion. Let me google that.

kkuepper: got it. I didn't realize that your suggestion actually was valid code. Changed it

components/track/TrackItem.vue
kkuepper: I also added another special case for when the album is known after your review.

PR #429, opened 2024-05-31
Feature/thumbnail buttons

cs1m0n:


electron/interface.d.ts
```diff
@@ -0,0 +1,9 @@
+export interface IElectronAPI {
+  setThumbarBtns: (mode: string) => Promise<void>;
```
kkuepper, line 2: can you change the name?
setThumbBarButtons

plugins/5.mediaKeys.ts
```diff
@@ -37,9 +37,11 @@ export default defineNuxtPlugin(() => {
         switch (state) {
           case MediaPlayerStatus.Paused:
             navigator.mediaSession.playbackState = "paused";
+            window.electronAPI.setThumbarBtns("paused");
```
kkuepper, line 40: in the browser electronAPI shouldn't exist. So we need to check if that object exists.
Maybe window.electronAPI?. is enough?

cs1m0n: Yes, tested it. That's enough.

electron/main.ts
```diff
@@ -40,6 +49,54 @@ const openWindow = (url: string) => {
     titleBarStyle: process.platform === "darwin" ? "hidden" : "default",
   });
 
+  const previousTrack = {
+    tooltip: "Previous",
```
kkuepper, line 53: Maybe we should translate this but it's probably difficult in this file.
Can you add a //ToDo: figure out how to translate the tooltip

kkuepper (approved):

PR #430, opened 2024-05-31
add link for admins

kkuepper:


components/ProfileMenu.vue
```diff
@@ -55,6 +56,14 @@ const saveAndCloseContentLanguageDialog = () => {
   )
     contentLanguageStore().contentLanguages = newLanguages;
 };
+
+const currentUser = useCurrentUser();
```
sifferhans, line 60: Isn't it better to directly use the reactive ref you get from the `useCurrentUser` composable?
Since it is reactive, the UI will update once the data is fetched.

```ts
const { data: user } = useCurrentUser()
```

sifferhans (changes requested):

sifferhans:
@kkuepper Are there plans to migrate the admin panel to the new app as well?

kkuepper:
> @kkuepper Are there plans to migrate the admin panel to the new app as well?

No. The old admin panel does an ok job and there’s no pressing reason to change it.
We might slowly move features into other systems though.

components/ProfileMenu.vue
kkuepper: that's a lot better. Thanks for pointing it out!

PR #432, opened 2024-06-03
don't modify external links

kkuepper:


SimonSimCity:
Don't know if there are other URLs like our dev-URLs, but looks good otherwise 😎👍

SimonSimCity (approved):

PR #434, opened 2024-06-04
Feature/428 fixes

kkuepper:
- add all search results to the queue (same as BMM app)
- remove duplicate contributors (e.g. track/108815)

### DIFF
--- a/utils/uniqueItems.ts
+++ b/utils/uniqueItems.ts
@@ -1,0 +1,9 @@
+import type { TrackModelContributorRelation } from "@bcc-code/bmm-sdk-fetch";
+
+export default function uniqueItems(
+  contributors: TrackModelContributorRelation[],
+) {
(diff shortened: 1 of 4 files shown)


utils/uniqueItems.ts
sifferhans, line 1: Couldn't we use a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) here?

sifferhans (approved):
Just a question about the `uniqueItems` function, other than that looks good

utils/uniqueItems.ts
kkuepper: I guess. But filter is working as well. So why change it after is has been implemented and tested?

sifferhans: Yeah, of course 👍🏼

PR #437, opened 2024-06-09
Queue: Add reorder functionality

### LINKED ISSUE #189: Queue: Allow the user to change the order of the items
Allow the user to change the order of the items in the queue. A plugin, which might be useful is https://github.com/SortableJS/vue.draggable.next

sifferhans:
Fixes #189 (?)

I used the [vue-draggable-plus](https://github.com/Alfred-Skyblue/vue-draggable-plus), as [vue-draggable-next](https://github.com/SortableJS/vue.draggable.next) doesn't have great TypeScript support, and seems not actively maintained.

https://github.com/bcc-code/bmm-web/assets/18753964/d8e72a9e-fa4d-4734-8a7b-07b907723106

### DIFF
--- a/components/media-player/MediaPlayerOpen.vue
+++ b/components/media-player/MediaPlayerOpen.vue
@@ -1,5 +1,7 @@
 <script setup lang="ts">
 import { MediaPlayerStatus } from "~/plugins/mediaPlayer/mediaPlayer";
+import { useDraggable } from "vue-draggable-plus";
+import type { TrackModel } from "@bcc-code/bmm-sdk-fetch";
 
 const { t } = useI18n();
 
@@ -24,7 +26,23 @@ const {
   rewind,
   fastForward,
   repeatStatus,
+  moveTrack
 } = useNuxtApp().$mediaPlayer;
+
+function isCurrentTrack(track: TrackModel) {
+  return track.id === currentTrack.value?.id;
+}
+
+const queueListElement = ref<HTMLUListElement>()
+
+useDraggable(queueListElement, queue, {
+  animation: 200,
+  onSort({ oldIndex, newIndex }) {
+    if(oldIndex === undefined || newIndex === undefined) return
+
+    moveTrack(oldIndex, newIndex)
+  }
+}) 
 </script>
 
 <template>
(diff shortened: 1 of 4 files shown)


plugins/mediaPlayer/mediaPlayer.ts
SimonSimCity, line 1: Can we move this functionality to the queue?

components/media-player/MediaPlayerOpen.vue
```diff
@@ -24,7 +26,23 @@ const {
   rewind,
   fastForward,
   repeatStatus,
+  moveTrack
 } = useNuxtApp().$mediaPlayer;
+
+function isCurrentTrack(track: TrackModel) {
```
SimonSimCity, line 32: This means that if I have a track in the queue twice, both will be marked as playing, or paused... Can we fix this?

SimonSimCity (changes requested):
Looks good apart from the two things mentioned.

SimonSimCity:
Just one last change ... can you ensure in your test that the prop `currentTrack` does not trigger as changed when moving the track? I just want to have this also as part of the spec of the added functionality.

This is important as the player will otherwise rebuild the track. This might not be the case now, but I just want this as part of the spec so we can ensure it stays like this in the future.

sifferhans:
@SimonSimCity Should be done now 👍🏼

SimonSimCity:
Perfect 🤩 Just the linter ...

SimonSimCity (approved):

components/media-player/MediaPlayerOpen.vue
```diff
@@ -198,19 +214,19 @@ const {
...
-        <li v-for="(item, i) in queue" :key="i" @click="queue.index = i">
+      <ul ref="queueListElement" class="px-3 pb-3">
+        <li v-for="(item, i) in queue" :key="item.id" @click="queue.index = i">
           <div
             :class="
-              queue.index === i ? 'bg-tint text-black-1 hover:bg-tint' : ''
+              isCurrentTrack(i) ? 'bg-tint text-black-1 hover:bg-tint' : ''
             "
-            class="flex cursor-pointer justify-between gap-2 rounded-xl px-3 py-2 transition-all duration-500 ease-out hover:bg-background-2"
+            class="flex cursor-row-resize justify-between gap-2 rounded-xl px-3 py-2 transition-all duration-500 ease-out hover:bg-background-2"
```
kkuepper, line 223: @sifferhans @SimonSimCity I think it makes sense to use a different mouse cursor to indicate that it's movable

kkuepper (approved):
This looks great! I love it!
@sifferhans Feel free to merge

components/media-player/MediaPlayerOpen.vue
sifferhans: Yeah, good point!

PR #441, opened 2024-06-12
Introduce hamburger menu & improve for small screen

kkuepper:


assets/icons/icon.hamburger.svg
sifferhans, line 1: Should probably be white when in dark mode for readability 😁

CleanShot 2024-06-13 at 00 05 20@2x

components/DocumentList.vue
sifferhans, line 1: Doesn't have anything to do with the mobile fixes, but nice either way 👍🏼

tailwind.config.ts
```diff
@@ -17,6 +17,7 @@ const config: Partial<Config> = {
         tracklist: "min-content auto auto auto min-content",
         tilesWide: "repeat(auto-fit, minmax(350px, 1fr))",
         tilesNarrow: "repeat(auto-fit, minmax(200px, 1fr))",
+        tilesOneLine: "repeat(6, minmax(220px, 1fr))",
```
sifferhans, line 20: Where do you get the number 6 from?

sifferhans (approved):

sifferhans:
Could also be nice with a simple transition for the mobile menu 😁 A slide-out or something like that

https://github.com/bcc-code/bmm-web/assets/18753964/64a91545-c01b-412b-a0ba-df3061c18dd2

assets/icons/icon.hamburger.svg
kkuepper: done

components/DocumentList.vue
kkuepper: I guess it's not a fix, but I think it's important to communicate that they should use the mobile app since this it provides a much better experience.

tailwind.config.ts
kkuepper: I don't see us ever having more than 6 tiles. It could just as well be 10.

kkuepper:
> Could also be nice with a simple transition for the mobile menu 😁 A slide-out or something like that

True. That would be nice.
Between 11% and 25% of visitors are from a mobile device.
Would you mind creating a PR for that? I think you can do that a lot faster. @sifferhans

PR #457, opened 2024-06-18
implement archive

kkuepper:


pages/archive/index.vue
```diff
@@ -0,0 +1,22 @@
...
+const { data: years } = useYearList();
+</script>
+
+<template>
+  <div>
+    <div class="h-3"></div>
+    <div class="columns-2 md:columns-3 xl:columns-4 2xl:columns-5">
+      <NuxtLink
+        v-for="(year, i) in years?.toReversed()"
+        :key="`year${i}`"
```
sifferhans, line 14: Using the index as the key here has no effect, as indexes might easily change (reordering, adding items, etc).
We should use some kind of unique identifier if possible. 

https://vuejs.org/guide/essentials/list.html#maintaining-state-with-key

sifferhans (changes requested):
Looks good! Only thing is that we should use unique values for `key` attributes in v-for loops. This is something we should do in general, not just here, but good to start somewhere 😁

pages/archive/index.vue
kkuepper: done

PR #460, opened 2024-06-19
extract TrackCollectionHeader and make it more responsive

kkuepper:


sifferhans (approved):
Nice! We could maybe go even further by wrapping the title down to the next line and making the cover take the full width on mobile screens?

Something along these lines:
CleanShot 2024-06-19 at 22 34 50@2x

PR #465, opened 2024-06-25
Profile menu: display signed-in user's information

sifferhans:
Closes #361 

CleanShot 2024-06-25 at 15 45 45@2x

PR #467, opened 2024-06-25
show who is logged in

kkuepper:


sifferhans:
I see we worked on the same thing 😅
Could you take a look at #465 ?

PR #470, opened 2024-07-04
provide playbackOrigin for CuratedPlaylist, Contributor and Tile

kkuepper:
So far this information has mainly been used for Analytics purposes. But we also use it to determine which are "Your top playlists" and we're getting reports about people listening to Daily Mix a lot, but not seeing it in their top playlists.

This PR creates the necessary infrastructure and sets the correct values for 3 of the 19 views. The other views can be added at a later time.

### DIFF
--- a/components/TileItem.vue
+++ b/components/TileItem.vue
@@ -7,12 +7,18 @@ const props = defineProps<{
 }>();
 
 const { setQueue } = useNuxtApp().$mediaPlayer;
+const origin = () => `Tile`;
 
 function playTrack() {
   if (!props.item.track) return;
   if (props.item.lastPositionInMs)
-    setQueue([props.item.track], 0, props.item.lastPositionInMs / 1000);
-  else setQueue([props.item.track]);
+    setQueue(
+      [props.item.track],
+      0,
+      origin(),
+      props.item.lastPositionInMs / 1000,
+    );
+  else setQueue([props.item.track], 0, origin());
   // ToDo: load linked album (from showAllLink) and add remaining items to the queue
 }
 
@@ -21,7 +27,7 @@ async function shufflePodcast() {
     const tracks = await new PodcastApi().podcastIdShuffleGet({
       id: props.item.shufflePodcastId,
     });
-    setQueue(tracks);
+    setQueue(tracks, 0, origin());
   }
 }
 </script>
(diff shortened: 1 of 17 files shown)


components/TileItem.vue
```diff
@@ -7,12 +7,18 @@ const props = defineProps<{
 }>();
 
 const { setQueue } = useNuxtApp().$mediaPlayer;
+const origin = () => `Tile`;
```
sifferhans, line 10: Is there a reason for this to be a function?

pages/playlist/curated/[id]/index.vue
```diff
@@ -10,15 +10,16 @@ const { data: playlist } = useCuratedPlaylist({ id: playlistId });
 const { data: tracks, pending } = useCuratedPlaylistTracks({ id: playlistId });
 
 const { setQueueShuffled, setQueue } = useNuxtApp().$mediaPlayer;
+const origin = () => `CuratedPlaylist|${playlistId}|${playlist.value?.title}`;
```
Comment (sifferhans) [inline pages/playlist/curated/[id]/index.vue:13]: These could probably be `computed` values, so Vue can optimize things better.

sifferhans (changes requested):
Nice 👌🏼 Just a small thought around the dynamically generated origins, but otherwise good

pages/playlist/curated/[id]/index.vue
Comment (sifferhans) [reply pages/playlist/curated/[id]/index.vue:13]: When using a `computed`, Vue will only re-run the getter if any reactive values inside it are updated. 
In this case it will only re-run when `playlist.value.title` is updated, and otherwise just return the previously computed value

components/TileItem.vue
kkuepper: I think Tile is the only one that doesn't have a dynamic value. So I thought it's more consistent. But it can just as well be constant.

pages/playlist/curated/[id]/index.vue
Comment (kkuepper) [reply pages/playlist/curated/[id]/index.vue]: done

sifferhans (approved):

SimonSimCity:
I'd personally prefer rather to extend the object rather than wrapping it. @kkuepper, was there a special reason for the class `EnrichedTrackModel` instead of a ts-type extension to `TrackModel`?

PR #478, opened 2024-08-26
feat(player): smoothly animate player width when toggling open state

sifferhans:
This is how it currently looks:

https://github.com/user-attachments/assets/6d73bc75-a968-46c2-8355-7bbeffc4fe96

With the transition it looks like this:

https://github.com/user-attachments/assets/b58471eb-ab9b-4aa2-8819-a469bd77a101


kkuepper (approved):
Looks pretty cool! Thanks a lot!

PR #481, opened 2024-09-02
feat: add sentry integration

sifferhans:
We are migrating to Sentry for error and crash reporting.


.github/workflows/build-and-deploy.yml
```diff
@@ -41,6 +41,7 @@ jobs:
         run: pnpm build
         env:
           NUXT_PUBLIC_APPLICATION_INSIGHTS: ${{ vars.NUXT_PUBLIC_APPLICATION_INSIGHTS }}
+          NUXT_PUBLIC_SENTRY_DSN: ${{ vars.NUXT_PUBLIC_SENTRY_DSN }}
```
sifferhans, line 44: This needs to be added to the repos Github vars

kkuepper (approved):
looks good to me. If you're confident that it will work as expected, you can merge it

PR #499, opened 2024-09-26
implement adding of album, private & curated playlists

kkuepper:


components/album/SubAlbum.vue
```diff
@@ -64,6 +64,8 @@ const onResume = () => {
       </div>
 
       <div v-if="active && album.children" class="ml-auto flex gap-2">
+        <AlbumMenu :album="album"></AlbumMenu>
+        <!-- ToDo: this is currently not clickable since we're not preventing event bubbling -->
```
kkuepper, line 68: @sifferhans When clicking the 3 dots, it closes the subalbum instead of opening the menu. Not sure how to prevent the event from bubbling. Can you have a look?
If it's not quickly possible, I'll just remove it for subalbums.

components/AddToPlaylist.vue
```diff
@@ -0,0 +1,73 @@
...
+const emit = defineEmits<{
+  close: [];
+}>();
+
+const { data: playlists } = usePrivatePlaylists();
+
+const selectList = async (playlistId: number, playlistName: string) => {
+  try {
+    const promise = props.listSelected(playlistId);
+    emit("close");
```
sifferhans, line 18: Isn't it better to close the dialog _after_ the promise has resolved?

components/AlbumMenu.vue
```diff
@@ -0,0 +1,63 @@
...
+
+const { t } = useI18n();
+
+const props = defineProps<{
+  album: AlbumModel;
+}>();
+
+const showAddToPlaylist = ref(false);
+
+const dropdownMenuItems = () => {
```
sifferhans, line 13: This should be a `computed` instead of a getter function

composables/playlist.ts
```diff
@@ -50,3 +50,32 @@ export function addTrackToPlaylist(playlistId: number, trackId: number) {
     link: [`</track/${trackId}>,`],
   });
 }
+
+export function addAlbumToPlaylist(playlistId: number, albumId: number) {
```
sifferhans, line 54: If we care about following the Nuxt conventions, these three functions are not composables,  but utils, and should be in `utils/playlist.ts`.

https://vuejs.org/guide/reusability/composables.html#what-is-a-composable

components/CuratedPlaylistMenu.vue
```diff
@@ -0,0 +1,63 @@
...
+
+const { t } = useI18n();
+
+const props = defineProps<{
+  playlist: PlaylistModel;
+}>();
+
+const showAddToPlaylist = ref(false);
+
+const dropdownMenuItems = () => {
```
sifferhans, line 13: This should be a `computed`

sifferhans (changes requested):
Some getter function that should be `computed`, but otherwise looks good 👍

I fixed event bubbling for the sub-album menu btw

composables/playlist.ts
kkuepper: we actually don't need those functions anymore

components/AddToPlaylist.vue
kkuepper: it currently works as an optimistic UI.

sifferhans (approved):

PR #502, opened 2024-09-30
Focus search input field on initialization

steffanhalv:


pages/search/[[term]].vue
```diff
@@ -75,6 +76,10 @@ onMounted(() => {
   isMounted.value = true;
   const main = ref<HTMLElement | null>(document.querySelector("main"));
 
+  if (searchbox.value) {
```
Comment (sifferhans) [inline pages/search/[[term]].vue:79]: This might be better to move out to a watcher instead of in `onMounted`, as the input might not exist yet here:

```ts
watch(searchbox, (box) => {
  if (box) {
    box.focus();
  }
});
```

sifferhans (changes requested):
Thanks for the PR! 🙌 
Things like this improves the user experience a whole lot

Just a small change needed, and it looks good to me!

sifferhans (approved):

PR #508, opened 2024-10-17
Feature: Fra Kåre dashboard

sifferhans:


sifferhans:
@kkuepper What do you think about the UX now?

locales/en.json
```diff
@@ -210,5 +210,15 @@
...
     }
+  },
+  "dashboards": {
+    "title": "Dashboard",
+    "fra-kaare": {
+      "title": "Fra Kåre",
+      "largeChurches": "Large Churches",
+      "smallChurches": "Small Churches",
+      "oneEpisodeDescription": "Listened to at least one episode",
+      "allEpisodesDescription": "Listened to every episode"
```
kkuepper, line 221: the description for one episode and all episodes will probably change relatively often. Therefore I'd prefer to not translate it.

pages/dashboards/fra-kaare.vue
```diff
@@ -0,0 +1,223 @@
...
+      : statistics.value?.smallChurches) ?? [],
+);
+
+function sortPercentageColumn(
+  a: number | null | undefined,
+  b: number | null | undefined,
+) {
+  if (typeof a !== "number" || typeof b !== "number") return 0;
+  const aInt = a * 100;
+  const bInt = b * 100;
```
kkuepper, line 45: I wonder if using int is not good enough for the change column

PR #526, opened 2024-11-08
introduce ApiPlaceholder

kkuepper:


sifferhans (approved):
lgmt. Is it to prevent hard-coding in the API endpoint?

PR #527, opened 2024-11-12
feat: lyrics page

sifferhans:


pages/lyrics/[id].vue
```diff
@@ -0,0 +1,221 @@
...
+});
+
+const saving = ref(false);
+async function saveLyrics() {
+  if (!lyrics.value) return;
+  saving.value = true;
+  try {
+    await new LyricsApi().lyricsIdPut({
+      id: Number(route.params.id),
+      lyrics: lyrics.value,
```
Comment (kkuepper) [inline pages/lyrics/[id].vue:64]: can you add Source: Manual?

pages/lyrics/[id].vue
```diff
@@ -0,0 +1,221 @@
...
+const lyrics = ref<Lyrics>();
+
+onMounted(async () => {
+  lyrics.value = await new LyricsApi().lyricsIdGet({
+    id: Number(route.params.id),
+  });
+
+  // Default longCopyright if not set
+  lyrics.value.longCopyright ??= `© Stiftelsen Skjulte Skatters Forlag, Norway. All rights reserved.
+forlaget@skjulte-skatter.no / activechristianity.org / christianbookshop.org`;
```
Comment (kkuepper) [inline pages/lyrics/[id].vue:16]: I think we should only have the copyright information. Not the promotion of some websites.

pages/lyrics/[id].vue
```diff
@@ -0,0 +1,221 @@
...
+
+const lyrics = ref<Lyrics>();
+
+onMounted(async () => {
+  lyrics.value = await new LyricsApi().lyricsIdGet({
+    id: Number(route.params.id),
+  });
+
+  // Default longCopyright if not set
+  lyrics.value.longCopyright ??= `© Stiftelsen Skjulte Skatters Forlag, Norway. All rights reserved.
```
Comment (kkuepper) [inline pages/lyrics/[id].vue:15]: setting that value should be done when sending to the server. Otherwise you have to select all text and replace it. Same applies to the verses as well.

components/LyricsEditor.vue
```diff
@@ -0,0 +1,83 @@
+<script setup lang="ts">
```
kkuepper, line 1: It currently sends html to the server. Which it should not.

kkuepper, line 1: For me, the editor didn't work very well for https://telegra.ph/Miracle-07-20

pages/lyrics/[id].vue
Comment (sifferhans) [reply pages/lyrics/[id].vue:15]: Ah, I see
So this shouldn't be visible in the UI at all before writing something manually? I guess we could use the `placeholder` attribute on the copyright textfield, at least

components/LyricsEditor.vue
sifferhans: Right, forgot to implement the cleanup

sifferhans: Could you try again now?

kkuepper: Works great now!

components/LyricsEditor.vue
```diff
@@ -0,0 +1,107 @@
...
```
kkuepper, line 1: The lyrics for miracle are pretty long and don't fit on a whole screen. Then I need to scroll up all the time to get to the button

components/LyricsEditor.vue
kkuepper, line 1: can you rename it to Text and Melody. (just in the UI)
And also have Text first?

kkuepper:
open tasks:
- make style look like in the app
- keyboard shortcut
- don't break paste
- sticky "verse title"
- copyright placeholder
- redirect on create

sifferhans:
@kkuepper Take a look now

components/AppToolbar.vue
```diff
@@ -1,9 +1,23 @@
 <script setup type="ts">
 const route = useRoute();
 const hasHistory = () => window.history.length > 1; // using a computed() doesn't work on inital load in the Electron app. No idea why. See PR #433
+
+onMounted(() => {
```
kkuepper, line 5: looks unfinished

components/LyricsEditor.vue
```diff
@@ -49,17 +50,6 @@ onMounted(() => {
       attributes: {
         class: "p-6 rounded-b-2xl",
       },
-      handlePaste(_, event) {
```
kkuepper: that was pretty useful. I don't think we should just remove it. But maybe there's a solution that doesn't break the default copy paste. E.g. a button to remove formattings.

components/LyricsEditor.vue
kkuepper, line 1: I forgot that Tobias asked for a field for Publishing year

components/AppToolbar.vue
sifferhans: This is for the lyrics editor, to make sure the "Verse title" button sticks to the bottom of the apps toolbar

kkuepper: but why do you call onMounted without any code in it?

sifferhans: Ah yes, that is wrong 😂 Will remove

kkuepper (approved):

PR #548, opened 2025-01-08
use fallback language for when translation is empty

kkuepper:


i18n.config.ts
```diff
@@ -35,4 +35,21 @@ export default defineI18nConfig(() => ({
     ro,
     ru,
   },
+  fallbackRootWithEmptyString: true,
+  postTranslation: (value, path) => {
+    if (value === "") {
+      const parts = path.split(".");
+      let message = en;
+      return "empty translation";
+      for (const key of parts) {
```
kkuepper, line 44: @sifferhans this fails if the message uses {placeholder}.
Have you solved this before?

sifferhans: I have not, no

i18n.config.ts
```diff
@@ -15,24 +17,48 @@ import pt from "./locales/pt.json";
...
+  Object.keys(obj).forEach((key) => {
+    if (typeof obj[key] === "object") {
+      obj[key] = cleanupEmptyProperties(obj[key], fallback[key]);
+    }
+    if (obj[key] === "") {
+      obj[key] = fallback[key];
+    }
+  });
+  return obj;
+}
```
sifferhans, line 30: We might want to check if the `obj` and `fallback` params actually are objects, as their types are `any` and thus the type system won't help us.

Not that important though, as the function is only used in this file

sifferhans (approved):
Nice 👍

PR #559, opened 2025-02-21
New translations for es committed.

BCC-Phrase:


sifferhans (changes requested):
The file is placed at the root of the project, not in `/locales`

Issue #219, opened 2023-09-20
Shuffle algorithm is not a proper shuffle algorithm


Ghostbird:
This is not a random shuffle algorithm:
https://github.com/bcc-code/bmm-web/blob/284ec84af2ad5d40fcd3059ac16a0a5510f49a77/plugins/mediaPlayer/Queue.ts#L47

### Proof

1. For an array of $n$ elements, there are exactly $n!$ permutations (i.e. possible shuffles).
2. Every comparison during a shuffle is a choice between two sets of permutations. For a random comparator, there is a $\frac{1}{2}$ chance of choosing each set.
3. For each permutation $p$, the chance of ending up with permutation $p$ is a fraction with denominator $2^k$ (for some $k$), because it is a sum of such fractions (e.g. $\frac{1}{8} + \frac{1}{16} = \frac{3}{16}$).
4. For $n = 3$, there are six equally-likely permutations. The chance of each permutation, then, is $\frac{1}{6}$. This can't be expressed as a fraction with a power of 2 as its denominator.
5. Therefore, the coin flip sort will never result in a fair distribution of shuffles.

### Example

A short visualisation of the problem would be this:

Divide a cake among six people[^1] but you can only do so by cutting a piece exactly into two equal halves[^2]. Either you'll end up with two extra pieces (eight equal pieces), or you'll end up with four smaller pieces and two bigger pieces[^3].

### Other issues

The sort is not well-defined, as it generates a random number for each _individual comparison_. You _ought to_ define a sort as completed when for every comparison between two adjacent elements, there is no reason to change the order. Mathematically the current sort algorithm should then run for an arbitrary, and probably long, amount of time. That is until the randomness coincidentally generates the same sequence twice in a row. This suggests that the actual result is heavily implementation dependent, since in practice, it runs much faster than mathematically proper.

## Proposed solution

Change it to a Schwartzian transform shuffle, I think the result will be just as good as [Fisher-Yates used by the BMM app](https://github.com/bcc-code/bmm-app/blob/a48eb919ea833299397b45314c4b23001a1ed733/BMM.Core/NewMediaPlayer/ShuffleableQueue.cs#L102), but I think this algorithm is one of the easiest to comprehend.

```ts
['A', 'B', 'C']
  // Wrap each entry and add a random sort number,
  // assume that the chance that two entries get exactly the same value is negligible
  .map((x) => ({ original: x, sort: Math.random() }))
  // Sort by the random numbers
[...]

PR #570, opened 2025-03-15
Replace shuffle algorithm

Ghostbird:
Previous algorithm had several issues. See #219 for the details.

New algorithm:
- Retains desired non-shuffle behaviour. The current track is kept at the top.
- Proper randomisation only dependent on the randomness of `Math.random()`, not on the pre-shuffle order of the track list.
- Proper stable comparison, the result of comparing two tracks A and B is the same at any point during the shuffle process.

I chose the Schwartzian Transform sort as basis for the shuffle algorithm. It is fairly easy to understand _why_ this algorithm is only dependent on the source of randomness. Compared to the BMM app's Fisher-Yates, which requires more mathematical proof that it is so, or the old implementation which required more mathematical proof that it is _not_ so.

### DIFF
--- a/plugins/mediaPlayer/Queue.ts
+++ b/plugins/mediaPlayer/Queue.ts
@@ -52,11 +52,22 @@ export default class Queue extends Array<EnrichedTrackModel> {
     if (this.sortedArray) return;
 
     this.sortedArray = [...this];
-    this.sort((a, b) => {
-      if (a === this.currentTrack) return -1;
-      if (b === this.currentTrack) return 1;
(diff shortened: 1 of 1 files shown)


sifferhans (approved):

sifferhans:
Works locally for me 🙌

Ghostbird:
> Looks good to me. Though I think the comments are a bit too wordy. But I think it's fine to merge. I'll just make it deploy and test it before merging.

I wondered about that too. My reasoning was that it would at least clarify the code for the review, and could always have been removed if desired.

PR #571, opened 2025-03-15
Update Vitest + VSCode extension

Ghostbird:
The recommended vitest plugin for VSCode was deprecated.
I installed the replacement instead. Then realised I needed to update Vitest to use that. Then realised I needed to update some devDependencies to run the newer vitest.

I also updated three tests where a call signature was used that was deprecated in Vitest 3 and would be removed in Vitest 4. The change is that previously the signature was `it('name', () => test, { options })` now it is `it('name', { options }, () => test)`.

The advantage of the new approach is that you can see the test name, and options at the start. The test function body might be large enough to push the options as third argument off-screen, even though they might be vital to understanding the test.


sifferhans (approved):

sifferhans:
Looks good, but there seems to be some linting issues. Could you take a look at it @Ghostbird?
Did you try to run the tests as well?

Ghostbird:
Ah, it actually crashes the linter. It might be that `eslint-plugin-vitest` is not compatible with ESLint 8.

I was working on a migration to ESLint 9 on another branch, but that is annoyingly complicated.
I'll see if I can downgrade `eslint-plugin-vitest` to an intermediate version that still works with ESLint 8.

sifferhans:
I can also take a look at the linter problem if you want 👍 @Ghostbird

sifferhans:
We should probably switch to `@vitest/eslint-plugin` and use `plugin:@vitest/legacy-recommended`. 
https://github.com/vitest-dev/eslint-plugin-vitest

sifferhans (approved):

PR #582, opened 2025-03-25
Playback Speed

### LINKED ISSUE #436: Playback speed control
I got feedback that someone missed the playback speed controls of the mobile app. Is this something that can be designed and implemented?

starkris51:
Requested feature from #436

Adds a new option in track menu to set playback speed in a song, only shown in media player and currently playing track

playback speed will be applied to a whole queue, when playing a new song the playback speed will reset

closes #436 

image
image


components/media-player/MediaPlayerOpen.vue
sifferhans, line 1: What is the reasoning for this change?

plugins/mediaPlayer/MediaTrack.ts
```diff
@@ -299,6 +302,10 @@ export default class MediaTrack {
     this.audioElement.volume = volume;
   }
 
+  setPlaybackRate(rate: number) {
+    this.audioElement.playbackRate = rate;
+  }
```
sifferhans, line 307: We could consider adding some tests for this

components/media-player/MediaPlayerOpen.vue
starkris51: to let the trackMenu know where the menu is opened from since playbackspeed should only show up in media player and menu for currently playing track

starkris51:
Not sure if the icon is right, there dosen't seem to be a icon for playback speed

sifferhans:
> Not sure if the icon is right, there dosen't seem to be a icon for playback speed

I think the icon is ok, as we don't have any design for the playback speed controls.

kkuepper:
@starkris51 thanks for committing this from a fork. It allowed us to realize that certain build steps aren't possible to run from a fork because secrets are not exposed on Pull requests from forks. I have included some changes on this PR, so that these build steps will be skipped in the future.
So thanks for your patience so that we could improve the process when committing from forks. Also, you now have write access on the bmm-web repository as well, so that you can assign issues yourself and create PRs not from a fork, if you want.

PR #600, opened 2025-04-03
Add simple client-side caching

sifferhans:
This pull request adds a quick and easy client-side caching solution using some of Nuxt's built-in functionality.

I've implemented a thin wrapper around `useLazyAsyncData` with an added `getCachedData` method. This method gets cached data from Nuxt's payload if the key already exists.

This solution is only per session though; it doesn't cache data for offline use.

### DIFF
--- a/composables/album.ts
+++ b/composables/album.ts
@@ -8,12 +8,12 @@ export function useAlbum(options: UseAlbumOptions) {
   const { id } = options;
 
   return reactiveApi(
-    useLazyAsyncData(`album-${id}`, () => new AlbumApi().albumIdGet({ id })),
+    useCachedLazyAsyncData(`album-${id}`, () => new AlbumApi().albumIdGet({ id })),
   );
 }
 
 export function useAlbums() {
   return reactiveApi(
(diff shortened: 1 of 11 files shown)


composables/album.ts
```diff
@@ -8,12 +9,24 @@ export function useAlbum(options: UseAlbumOptions) {
   const { id } = options;
 
   return reactiveApi(
-    useLazyAsyncData(`album-${id}`, () => new AlbumApi().albumIdGet({ id })),
+    useLazyAsyncData<AlbumModel>(
```
kkuepper, line 12: I would call this a very wordy solution, since it adds multiple lines of code in a lot of places. We should chat if this is the best solution.

kkuepper:
In Europe and North America the website feels responsive even without caching and this Pull Request therefore doesn't have any real improvement. On the other side it introduces cache invalidation problems and a solution that is harder to maintain. (e.g. the home page has streak information which changes regularly)
Because of those reasons, we're not merging this PR and we might revisit if we learn about a concrete case where caching is solving an acute problem.

PR #625, opened 2025-05-23
feat(streak): use point colors if defined

sifferhans:


components/StreakItem.vue
```diff
@@ -49,7 +53,7 @@ const getDays = (item: CurrentWeeksStreakVm) => [
       >
         <div
           class="aspect-square w-3 rounded-full"
-          :class="day.listened ? 'bg-[#DBE459]' : ' border border-label-4'"
+          :class="{ 'border border-label-4': !day.listened }"
```
sifferhans, line 56: The `bg-[#DBE459]` class was always overridden

kkuepper (approved):

PR #627, opened 2025-05-23
feat(streak): add overview dialog

sifferhans:
CleanShot 2025-05-23 at 15 43 32


kkuepper (approved):

kkuepper:
I think "time left to listen" is not vital for this feature to make sense. I created #629 for that.

PR #644, opened 2025-06-06
add additional statistics & make church selectable

kkuepper:


sifferhans:
@kkuepper I see that the selected church stats are the same for all churches. That seems to be a bug

PR #671, opened 2026-03-24
fix: prevent stale data in TrackMenu by forcing remount on track change

andreasgangso:
Using :key on the whole menu prevents future similar issues and is a bit easier to manage, but I'd also recommend updating to a newer nuxt version where you can pass a reactive key to useLazyAsyncData (so you could do `useTrackTranscription(() => props.track.id)` instead)


sifferhans (approved):
Looks good, thanks!

